PR #3, opened 2017-01-05
Added failing test with complex strings

rtsao:
It appears stylis doesn't currently handle tokens within strings correctly

```
.user .foo:before {"}.user .foo:before content: ".hello{world}
expected:
.user .foo:before {content: ".hello {world}"}
```


thysultan:
Thanks @rtsao, added a patch for this in v0.9.0 (haven't yet published to npm) - [CHANGELOG](https://github.com/thysultan/stylis.js/blob/master/CHANGELOG.md)

rtsao:
Awesome!

Issue #11, opened 2017-03-29
::placeholder not vendor prefixed


jaydenseric:
`::placeholder` is not vendor prefixed: [caniuse](http://caniuse.com/#feat=css-placeholder).

thysultan:
Thanks for opening this, have considered this before and forgot about it, do you know if this can be used as an element selector across browser implementations? i.e 

```scss
input.red:placeholder, 
input.error:placeholder {
    color:red;
}
```

jaydenseric:
It should be double colon because it's a pseudo element.

You can not mixes vendor prefixes in one selector; the rules have to be duplicated.

In:

```css
input.red::placeholder, 
input.error::placeholder {
  color:red;
}
```

Out:

```css
input.red::-webkit-input-placeholder, 
input.error::-webkit-input-placeholder {
  color:red;
}
input.red:-ms-input-placeholder, 
input.error:-ms-input-placeholder {
  color:red;
}
input.red::placeholder, 
input.error::placeholder {
  color:red;
}
```

thysultan:
added in v2.0.0

evenchange4:
I create a PR #13 for nested use case:

In:

```css
input {
  &::placeholder {
    color:red;
  }
}
```

Out:

```css
input::-webkit-input-placeholder {
  color:red;
}
input::-moz-placeholderr {
  color:red;
}
input:-ms-input-placeholder {
  color:red;
}
input::placeholder {
  color:red;
}
```

thysultan:
@evenchange4 Thanks for the PR, pushed v2.0.1 with the patch.

jaydenseric:
Thanks guys :)

PR #13, opened 2017-04-05
fix(placeholder prefix): should work with &::placeholder case

evenchange4:
for #11


thysultan:
@evenchange4 Thank you.

PR #29, opened 2017-06-02
fix a typo

dnakov:
flex is not defined


thysultan:
@dnakov Thanks, published v3.0.15

PR #33, opened 2017-06-06
Allow the colon to be the last character on the line without inserting the semicolon

fatfisz:
I found out that my styles have broken after last Stylis upgrades (I'm using styled-jsx):

```css
html {
  background-image:
    linear-gradient(0deg, rgba(255, 255, 255, 0.8), rgba(255, 255, 255, 0.8)),
    url(/static/background.svg);
}
```

A semicolon was inserted after `background-image:`, which resulted in the rule being broken and having no background on my website 😢 I double checked and the original rule is valid CSS.

### DIFF
--- a/stylis.js
+++ b/stylis.js
@@ -377,6 +377,7 @@
 						// valid characters that
 						// may precede a newline
 						switch (tail) {
+							case COLON:
 							case COMMA:
 							case NULL:
 							case TAB:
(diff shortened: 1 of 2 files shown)


thysultan:
Thanks @fatfisz

fatfisz:
Thanks for the quick merge! As soon as you release a new version I'll open a PR in styled-jsx repo.

thysultan:
@fatfisz pushed 3.0.17

PR #45, opened 2017-07-04
s/reversed/reverse

cheeaun:
[`animation-direction`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction) accepts `reverse` as one of its values, not `reversed` (with 'd'). This PR fixes this.


thysultan:
Thanks.

Issue #53, opened 2017-08-24
Position fixed with wrong prefixing


evenchange4:
There are useless prefixing of `position: fixed` when I update from `3.2.9` to `3.2.10`:

```diff
// compare 3.2.9 with 3.2.10

+ position: -webkit-box;
+ position: -webkit-fixed;
+ position: -ms-fixedbox;
  position: fixed;
```

Online demo:

![2017-08-24 11 00 59](https://user-images.githubusercontent.com/1527371/29647424-a8d95db6-88bb-11e7-823a-281f5179ccfd.png)

thysultan:
Pushed the patch in 3.2.11

PR #54, opened 2017-08-24
fix: position fixed without prefixing

evenchange4:
for #53


thysultan:
Thanks for finding this.

thysultan:
Published 3.2.11 with this patch.

PR #56, opened 2017-08-29
Add failing test for shorthand properties over multiple lines

benhjames:
Currently when writing something like:

```
background-image: linear-gradient(0deg, red, blue) center
  center no-repeat;
```

It gets interpreted as though `center` is its own property:

```
background-image:linear-gradient(0deg,red,blue) center;center:no-repeat;
```

The CSS spec does seem to allow breaking shorthands over multiple lines.

Not sure on best way to fix, so this PR is simply a failing test as an example. I imagine this would also happen for other shorthand properties like animations/transitions!

### DIFF
--- a/tests/spec.js
+++ b/tests/spec.js
@@ -1308,6 +1308,16 @@ var spec = {
 		expected: ``+
 		`.user html{background-image: linear-gradient(0deg,rgba(255,255,255,0.8),rgba(255,255,255,0.8)), url(/static/background.svg);}`
 	},
+	'multiline declaration with subproperties': {
+		sample: `
+			html {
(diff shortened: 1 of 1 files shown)


thysultan:
This is due to having the `no-semicolon` mode on. With the explicit semicolon mode this should work. Though i do think there should be a way to get this working for the no-semicolon option as well.

The way the `no-semicolon` would work is in the absence of a semicolon before a newline it waits to find the next non-whitespace/comma character to determine weather it's a property or otherwise; We might be able to extend this even further to when the next valid token that confirms it is not a new rule i.e colon/semicolon.

thysultan:
Patched in 3.4.5

PR #66, opened 2017-10-14
Fix @import in rule-sheet

mitchellhamilton:
This is just like 8e9a2ce but for `@import`, otherwise `@import` appears in another rule.


thysultan:
Thanks, published 0.0.6 with this patch.

Issue #63, opened 2017-10-11
Remove 2009 flex box spec


tkh44:
The spec is ancient and the prefixes are hurting performance for any browser that is ie11 and above.

> New syntax support is: Chrome 21+, Safari 6.1+, Firefox 22+, Opera (& Opera Mobile) 12.1+, IE 11+, and Blackberry 10+. 
    
(https://css-tricks.com/old-flexbox-and-new-flexbox/)

> Summary: Old is 2.3x slower than new.

   https://developers.google.com/web/updates/2013/10/Flexbox-layout-isn-t-slow
   https://developers.google.com/web/tools/lighthouse/audits/old-flexbox

Regardless of how modern browsers are using the older properties, we are sending down unnecessary properties when using SSR.

![pasted_image_at_2017_10_10_05_57_pm_360](https://user-images.githubusercontent.com/662750/31456852-37188fb6-ae78-11e7-9a63-de2ae919df85.png)

I think we should remove this for the benefit of all end users. I'm willing to help with the removal if you are open to the idea of removing the old spec.

tkh44:
cc @mxstbr @giuseppeg

thysultan:
I'm open to removing the non IE10+ variants if caniuse stats paint a good picture for < Chrome 21 and < Safari 6.1 which is mostly out of concern for varying android versions that could be using older versions of Chrome as defaults.

>The spec is ancient and the prefixes are hurting performance for any browser that is ie11 and above.

Does this still apply here? Since it ships the new syntax last, the browser should be using that instead right?

mxstbr:
`styled-components` supports all browsers React supports, which [according to the docs](https://reactjs.org/docs/react-dom.html#browser-support) means:

> React supports all popular browsers, including Internet Explorer 9 and above.

I'd be against removing anything that breaks applications in browsers that React supports. /cc @gaearon

giuseppeg:
As much as I would like to see legacy browsers die, I have second @mxstbr's comment. 
Styled-jsx users have the option to disable Stylis' auto prefixing feature and use autoprefixer via plugins.

@thysultan would it be possible to abstract the auto prefix logic into a separate plugin? In that case @tkh44 could write his own version / fork.

mxstbr:
Maybe it'd be possible to write a plugin that disables that part of the autoprefixing or something?

thysultan:
[...]

PR #67, opened 2017-10-14
3.4.0

thysultan:
This minor adds:

- dynamic vendor prefixing.

```js
stylis.set({
prefix: (key, value, context) => {
    return false // determine if you want to vendor prefix the rule 
}
})
```


thysultan:
The README updates demonstrate what it would look like. 

- context is equivalent to context in plugins, which allows you to also opt out of non property-value vendor prefixes like `:read-only` and `@keyframes`. 

- `key` and `value` are separate as opposed to the proposal in #63 to allow you to opt out of vendor prefixes that are value oriented like `display: flex;`

cc @tkh44 @mxstbr @giuseppeg

PR #78, opened 2017-12-20
fix vendor prefixing of align-content for IE10

### LINKED ISSUE #77: More IE flex things
`align-content: flex-start` -> `-ms-flex-line-pack: start`
`align-content: flex-end` -> `-ms-flex-line-pack: end`

~~`align-self: flex-start` -> `-ms-flex-item-align: start`
`align-self: flex-end` -> `-ms-flex-item-align: end`~~

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Advanced_layouts_with_flexbox

probablyup:
fixes #77

### DIFF
--- a/tests/spec.js
+++ b/tests/spec.js
@@ -567,6 +567,11 @@ var spec = {
 			}
 
 			div {
+				align-content:flex-start;
+				align-content:flex-end;
+			}
+
+			div {
 				color:papayawhip;
 				order:flex;
 			}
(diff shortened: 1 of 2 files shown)


tests/spec.js
```diff
@@ -566,6 +566,10 @@ var spec = {
 				align-content:value;
 			}
 
+			div {
+				align-content:flex-start;
```
thysultan: Lets add a test for `flex-end` as well.

probablyup: done

Issue #79, opened 2017-12-21
Bytes savings


Andarist:
I was playing around and I've wanted to check what would change if stylis would have constants inlined.

I've started from checking what is the actual gzipped size of the minified bundle and it is 4559. I don't know how to generate it, because you don't have any script in the repo for this task. I thought you were using google closure compiler so I've tried to use google-closure-compiler-js, but unfortunately it crashes (or I dont know how to use it) - "FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory".

So I've used something that I know, uglifyjs@3, minified the stylis and checked its gzipped size to have some point of reference. It got slightly better result (although I dont know what I am comparing with really) - 4548.

I've inlined manually [all char codes and special identifiers](https://github.com/thysultan/stylis.js/blob/4cc510b4069993e3a5d01fe5c241a3a28f094b70/stylis.js#L89-L127) through out the file and run uglifyJS agains and checked the gzipped size again, which was 4353.

That's not bad, nearly 200 bytes of savings. And of course the inlining could be automated with Babel or something as compilation step.

I've also wanted to check benchmarks to see if that helps anything (I suppose it helps a little or does no change), but PostCSS benchmarks are crashing on stylis at the moment so I couldn't check it reliably.

Would you be interested in following this and possibly having me implementing this constants inlining through with a babel?

mitchellhamilton:
Just wanted to mention that emotion ships a version of stylis that we run through closure compiler. I use the [online closure compiler](https://closure-compiler.appspot.com/home), enable advanced, paste in stylis and remove the last bit of the compiled code that is related to UMD. [It looks like this](https://github.com/emotion-js/emotion/blob/master/packages/emotion-utils/src/stylis.js)

thysultan:
I used uglifyjs for the current minified distribution artifacts in the repo.

I haven't tested with the JavaScript version of closure compiler but i think the Java [version](https://closure-compiler.appspot.com/home) does this out of the box. which is one of the reasons the jsdocs and bracket notations where used – motivated by the need to support both the simple and advanced mode of closure compiler(java version).

Do you have a reproduction of the crashes on google-closure-compiler-js? Is it specific to only stylis?

Andarist:
> Just wanted to mention that emotion ships a version of stylis that we run through closure compiler. I use the online closure compiler, enable advanced, paste in stylis and remove the last bit of the compiled code that is related to UMD. It looks like this



[...]

PR #87, opened 2018-01-08
Ship multiple module formats

### LINKED ISSUE #81: Ship es modules to npm
Would like to continue the issue mentioned in https://github.com/thysultan/stylis.js/issues/79

> That's actually another topic that I would like to tackle here - I think stylis should ship es & cjs bundles (maybe event closure compiled?) to npm, so libraries like emotion could consume it gracefully without having to copy its source.

The whole thing of shipping es modules, cjs modules and umd files is pretty easy. Stylis structure is somewhat special though and its the reason why I havent yet send PR for this. I would like to discuss what we can do to tackle this.

What is the problem?
Stylis references the factory closure, so it can act both as constructor and as a singleton. While I really am impressed by this cleverness (it took me few moments to see what's going on) it doesn't play well with existing tooling. Normally the umd factory is just a factory that creates a module and is not referenced from the inner closure.

In my opinion it would be better to just export something like:
```js
export function Stylis() { /* ... */ }
export default /*#__PURE__*/new Stylis()
```
That way the default export could even be removed by UglifyJS (but that's really just a single line of code and a single function call) if the consumer would be interested solely in the constructor style.

I'm also thinking that both `use` and `set` functions are technically tree-shakeable, but with the current approach of mixed singleton/constructors I have no idea how to support tree-shaking of those in a sane manner.

Let me know what do you think, I'm ready to invest time to refactor in direction that we agree upon.

Andarist:
fixes #81

Just a proof of concept at the moment. Please take a look at the output, seems ok for me, but I still need to test it thoroughly.

Having a build pipeline we could explore later including further compilation optimizations, such as constants inlining mentioned in #79, or even shipping to npm google closure compiled files (with source maps).


Andarist:
@thysultan friendly 🏓 😉

thysultan:
>The somewhat bigger gains are from having module entry point for stylis.

What are the gains to be had from having a module entry point?

Andarist:
Everything boils down to less bytes (albeit ofc not that much less) being shipped to the browsers.

For starters we could remove umd wrapper which costs some bytes and that would not have to be included in final applications.

In general module entry (written using ESM modules) allows bundlers for 2 size-optimization techniques:
- tree-shaking - which is **still relevant** here. If `stylis` is used by a dependency A which in turn stay unused more static nature of `module` entries allow bundlers to remove both unused dependency A and its dependencies (in this example `stylis`) from the final bundle. For umd/cjs format bundlers such as webpack just bails out of this optimization
- scope hoisting a.k.a. module concatenation - historically webpack (and other bundlers) was wrapping each module/file in a closure (similar to umd wrapper) and register such closure/module in its runtime registry under some id. Nowadays thanks to more static nature of ESM modules it tries to concatenate all modules into a single scope, but it bails out of this optimization for umd/cjs formats. So the gain here would be allowing bundlers to skip this module wrapping which costs around 17 bytes - which is ofc somewhat micro optimization, but having large part of the ecosystem still written using commonjs it all adds up

I also think that stylis not shipping a module format (or optimized source with closure compiler) is a primary reason for libraries like `emotion` to just copy its source code in their codebases instead of keeping it as a dependency.

Andarist:
Friendly 🏓 😉

rollup.config.js
```diff
@@ -0,0 +1,48 @@
+import uglify from 'rollup-plugin-uglify'
```
TrySound: `import { uglify } from 'rollup-plugin-uglify'`

Andarist: This branch uses [`rollup-plugin-uglify@2.0.1`](https://github.com/thysultan/stylis.js/pull/87/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R40), so it's good. If we get to move this forward then I'm gonna rebase & update deps.

thysultan:
## scope hoisting

I can see the immediate value preposition of this point. 

## tree-shaking

Where you able to test if the this point does indeed remove unused imports with this. I have a feeling the current factory closure approach would be a hard blocker for tree shaking at that level.

The next major release could improve on this and other API decisions.  The question to answer then is wether there is an API pattern pattern we could use that allows you to tree-shake the functions `isolate`, `proxy`, `minify`, `use`, `set` and a good chunk of the prefixing going on in `property` if you make use non of them publicly. 

>primary reason for libraries like emotion to just copy its source code

In an effort to crunch bytes i think emotion also runs the code through a JavaScript parser and removes some parts manually.

Andarist:
> Where you able to test if this point does indeed remove unused imports with this. I have a feeling the current factory closure approach would be a hard blocker for tree shaking at that level.

Yeah, it's not possible to tree-shake unused parts now. It only matters for the cases I've described - when `stylis` is used by A which we import in out our app somewhere, but A stays unused. With current setup A might get tree-shaken but `stylis` won't although it was only used by A.

> The question to answer then is wether there is an API pattern pattern we could use that allows you to tree-shake the functions isolate, proxy, minify, use, set and a good chunk of the prefixing going on in property if you make no use of them.

I don't think it's possible. It would only be possible if using sophisticated static analysis rules, **maybe** closure compiler would be able to do it for some cases, or `prepack` in the future. Don't know. Webpack's tree-shaking is rather simple algorithm (with numerous flaws/bailouts at the moment) and won't be able to handle such cases.

"Problem" with stylis is that it can be used both as a singleton and a factory. In theory you could export 2 separate builds - one for a singleton (default) and one for a factory (smth like `stylis/factory`). That way singleton entry would be able to tree-shake unused parts, but wouldn't be able to share code with `stylis/factory` - which is a downside of this approach. Personally I think it outweighs the benefit.

> In an effort to crunch bytes i think emotion also runs the code through a JavaScript parser and removes some parts manually.

According to the comment by @mitchellhamilton that I've read, he just:
- copies the code
- removes UMD wrapper
- runs through closure compiler
- adds `export default`
- saves the work

By doing most of those in the `stylis` itself emotion could just use `stylis` as dependency.

thysultan:
I think if we can keep the current file structure and add distribution artifacts  for`stylis.es.js` and `stylis.es.min.js` we could ship this as a non breaking change.

Though i'm unsure if adding these to the `pkg.module` field break webpack imports?

TrySound:
@thysultan I would call them more clearer `esm` like [esm loader](https://github.com/standard-things/esm). This name will replace `es` format in rollup one day (already available as alias).

TrySound:
Hi @thysultan Any news on this?

Andarist:
This PR is no longer needed, upcoming v4 fixes this problem.

PR #90, opened 2018-01-10
Workaround of Closure Compiler problem

maciej-ka:
Related to https://github.com/styled-components/styled-components/pull/1409

This is a workaround for a weird behavior of Closure Compiler, which is failing to minify Styled Components.

I don't know why this small move of variable helps. This PR is a result of experiments with commenting out fragments and trying different changes that have no effect on the logic of Stylis. Hopefully, Closure Compiler devs will figure out the root problem.


thysultan:
Can you open an issue with Closure Compiler so they know about this problem. It would be preferable to land a patch there.

maciej-ka:
The issue is already there: https://github.com/google/closure-compiler-js/issues/23.

thysultan:
I'm hesitant to introduce this workaround. The best course of action would be to open a new issue specific to this problem with Closure Compiler.

A minimal repro might entail some small module using stylis, and see if the above still registers as a workaround – this might help narrow down where the problem is in GCC.

wardpeet:
@maciej-ka any idea why this fixes it? If so I could have a look at closure compiler source and poke some people about it.

maciej-ka:
No idea why this fix is working.

I don't plan to spend more time on this. It took hours to blindly guess this fix and extracting minimal repro from [seven hundred lines parser function](https://github.com/thysultan/stylis.js/blob/master/stylis.js#L177-L884) makes my corporate work look interesting.

It's hard to say that and I don't mean to be rude. I hope honest statement is better here than giving an impression that progress is on the way.

Issue #64, opened 2017-10-12
Prepending & to nested pseudo selectors


seanevd:
I'm not sure if this is a bug, but I thought I'd ask since I saw this in some code I was reviewing the other day. This is styled-components. It compiles out correctly, but I didn't see it in the docs, anywhere.
```
const example = styled.div`
  color: green
  ::after {
    background: yellow;
  }
`;
```
The thing that is strange is that ::after pseudo selector does not have an & prepending it to the selector. Is this expected behavior, I don't think it works in scss (not sure if that's something you compare against). Just FYI.

thysultan:
Your right, it won't work in [scss](https://www.sassmeister.com/gist/a00d32b8ac6c0cdec706f595d63b0b63), however it is an intentional behavior here. Does the current behavior break something in your app?

seanevd:
No, it doesn't break anything, it's working fine. I'm just a long time Sass user and thought it would break or not render correctly. Why is it intentional? Is there somewhere I could read more about this?

thysultan:
>Why is it intentional?

Mainly DX, given that orphaned `:active{}` pseudo selectors are invalid css – stylis would use the surrounding namespace similar to how it might fix other invalid css patterns, i.e missing colons/semicolons etc.

elektronik2k5:
I too thought it's a bug. Should be documented, as it isn't what preprocessor users are used to. @thysultan would you like me to make a PR?

thysultan:
@elektronik2k5 Yes, that would be great.

bensaufley:
@thysultan I'm trying to find where this is explicitly stated:

> given that orphaned `:active{}` pseudo selectors are invalid css

Can you point me somewhere? I'm looking at the [Selectors Level 3 spec](https://drafts.csswg.org/selectors-3/#pseudo-classes) and I don't see anything but I might be missing it. It _seems_ like this says otherwise (emphasis mine):

> Pseudo-classes are allowed anywhere in sequences of simple selectors, after the leading type selector or universal selector **(possibly omitted).**

It's not a rule I was aware of, and I feel like standalone pseudo-classes have worked for me in the past. For what it's worth, [this codepen with a bunch of orphaned pseudo-classes and pseudo-elements](https://codepen.io/bensaufley/pen/GXpEaq) works for me on MacOS Chrome 68, Firefox 61, and Safari 11.1.2; as well as Edge 17 and IE11 on Windows 10.

[...]

PR #113, opened 2018-07-27
Don't remove whitespace before ::placeholder

Darep:
Adds special handling for CSS like `div.input ::placeholder { ... }` when `cascade` is set to false.

We are using a UI library which defines the placeholder CSS like that and noticed that the whitespace before the "::placeholder"-selectors was being removed and the placeholders looked wrong. I tracked this down from next.js => styled-jsx => stylis and noticed it only happens when `cascade` is false, which styled-jsx always uses.

Took a good few hours to figure out the fix this, but I feel like my solution might not be ideal :) Let me know what you think!

### DIFF
--- a/stylis.js
+++ b/stylis.js
@@ -821,6 +821,10 @@
 								}
 								case TAB:
 								case SPACE: {
+									// check for "::placeholder" after space and don't remove the space
+									if (body.charCodeAt(caret+1) === COLON && body.charCodeAt(caret+2) === COLON && body.charCodeAt(caret+3) === PLACEHOLDER) {
+										break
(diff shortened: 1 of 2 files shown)


bensaufley:
I think this is part of a bigger regarding all pseudo-selectors, discussed in https://github.com/thysultan/stylis.js/issues/64

Andarist:
As mentioned by @bensaufley - this was logic implemented in v3 for all pseudo-selectors. Changing it within v3 would be a major breaking change, but v4 is around the corner and it changes this behavior to what you have expected.

PR #122, opened 2018-09-20
plugins(custom-property): add fallback support

### LINKED ISSUE #121: custom-properties: don't overwrite, fall-back.
Hi!

Currently, the custom-properties plugin will transform a custom-property statically. I think that possibly a more appropriate behavior would be to fall-back.

https://github.com/thysultan/stylis.js/blob/master/plugins/custom-properties/index.js#L36

for example:

## INPUT
```js
import stylis from "stylis";
import cp from "stylis-custom-properties";

stylis.use(cp);

return stylis(
  "#id",
  `
:global(:root) {--fs: 1em}
font-size: var(--fs, 10px);
`
);
```

## ACTUAL
```css
#id{
  font-size:1em;
}
```

## EXPECTED
```css
:root{
 --fs: 1em;
}
#id{
  font-size:1em;
  font-size:var(--fs, 10px);
}
```

this is [how postcss-custom-properties works](https://github.com/postcss/postcss-custom-properties#postcss-custom-properties-) and I think is what many would expect. Would you be open to a PR introducing this functionality?

VinSpee:
instead of brute-force replacing the custom property with the defined
value, add two properties, one that retains the custom property, and one
fallback that contains the calculated value.

resolves #121

```node
❯ node
> const stylis = require('./stylis.js')
undefined
> const cp = require('./plugins/custom-properties/
./plugins/custom-properties/package  ./plugins/custom-properties/tests/

> const cp = require('./plugins/custom-properties/package
./plugins/custom-properties/package

> const cp = require('./plugins/custom-properties/package
./plugins/custom-properties/package

> const cp = require('./plugins/custom-properties/package
./plugins/custom-properties/package

> const cp = require('./plugins/custom-properties/package
./plugins/custom-properties/package

> const cp = require('./plugins/custom-properties')
undefined
> stylis.use(cp)
[Function: use]
> const style = stylis(
...   ".test",
...   `
... --fs: 2rem;
... font-size: var(--fs);
... `
... );
undefined
> style
'.test{--fs:2rem;font-size:2rem;font-size:var(--fs);}'
```


VinSpee:
Hi @thysultan, anything I need to do for you to accept this?

Thanks!

VinSpee:
Thanks! I'd like to expand support in the future to handle [custom property fallback values](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables#Custom_Property_fallback_values), but I figured this was a good start!

Issue #130, opened 2018-11-21
cursor: image-set() not being prefixed


1000hz:
Similar to #105, `image-set` is not being given the `-webkit-` prefix when used inside a `cursor:` declaration.

The following modified spec fails:
```js
	'vendor prefixing VIII': {
		sample: `
			background:image-set(url(foo.jpg) 2x);
			background-image:image-set(url(foo.jpg) 2x);
			cursor:image-set(url(foo.jpg) 2x), grab;
		`,
		expected: ``+
		`.user{`+
			`background:-webkit-image-set(url(foo.jpg) 2x);`+
			`background:image-set(url(foo.jpg) 2x);`+
			`background-image:-webkit-image-set(url(foo.jpg) 2x);`+
			`background-image:image-set(url(foo.jpg) 2x);`+
			`cursor:-webkit-image-set(url(foo.jpg) 2x), -webkit-grab;`+
			`cursor:image-set(url(foo.jpg) 2x), -moz-grab;`+
			`cursor:image-set(url(foo.jpg) 2x), grab;`+
		`}`
	},
```

thysultan:
Yes, we can add this. Do you want to try contributing this patch? It should be similar to the one listed for background-image.

1000hz:
I had tried when I filed the issue and got it working for everything but the `cursor:image-set(url(foo.jpg) 2x), grab;` case where it needs to prefix both `image-set` and `grab` in the same rule.   I've since deleted my copy of the repo. 😞

I'll try to redo what I had and send in a PR and we can work on fixing that failing case.

thysultan:
Sounds good 👍

Andarist:
@thysultan do we want to support prefixing this in v4?

thysultan:
Yes, it's on the backburner of PR's to review for V4

Andarist:
This got fixed on the v4 branch.

PR #135, opened 2018-12-28
Fix cursor:image-set() prefixing

1000hz:
Fixes the issue in #130.

The behavior before was testing for a `cursor` value of  `grab`/`zoom` and then prefixing whatever came after the colon. This resulted in `cursor: image-set(foo.jpg, 2x), grab` becoming
```css
cursor: -webkit-image-set(foo.jpg, 2x), grab;
cursor: -moz-image-set(foo.jpg, 2x), grab;
cursor: image-set(foo.jpg, 2x), grab;
```

which is incorrect.

This patch changes the test to add an additional check for image-set, and if found prefixes both image-set and zoom/grab properly, otherwise just checks for zoom/grab.


Andarist:
This got fixed on the v4 branch.

PR #148, opened 2019-06-20
Fix bug with @-ms-viewport

cameron-martin:
Currently `@-ms-viewport` gets the selector nested inside of it, i.e.

```
@-ms-viewport { width: device-width; }
```
compiles to 

```
@-ms-viewport { .selector { width: device-width; } }
```

Moreover, if the selector is empty then it compiles to invalid CSS:

```
@-ms-viewport { { width: device-width; } }
```

This PR adds failing test cases for this. However, I could not see how to fix it. Hopefully somebody who knows the codebase a bit better can contribute this part.

### DIFF
--- a/tests/spec.js
+++ b/tests/spec.js
@@ -62,9 +62,25 @@ var spec = {
 		`.user body{color:red;}`+
 		'}'+
 		`@viewport{min-width:640px;max-width:800px;}`+
+		`@-ms-viewport{width: device-width;}`+
 		`@counter-style list{system:fixed;symbols:url();suffix:" ";}`+
 		`@-moz-document url-prefix(){.user .selector{color:lime;}}`+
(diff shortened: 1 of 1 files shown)


cameron-martin:
@thysultan Any pointers on where to start looking to fix this?

thysultan:
We do something similar for @font-face and other "flat" at rules.

Andarist:
@thysultan do we want to support this non-standard rule in v4? I've checked and we have a similar issue when parsing `@-webkit-keyframes` because of those two hitting currently [this case](https://github.com/thysultan/stylis.js/blob/1acdcd9df11facf0fb31fcfc44cfee1bf7fa9c9b/src/Parser.js#L78) rather than the default one

thysultan:
Fixed in https://github.com/thysultan/stylis.js/commit/1a852159966f7078b43e0ae0918526425f8f0d6f

Issue #146, opened 2019-05-27
'default' is not exported by ../node_modules/stylis/stylis.min.js


wanbinkimoon:
As described in the title I'm not able to import the stylis constructor in any way, here is some code 

```javascript
import Stylis from 'stylis/stylis.min';

const stylis = new Stylis({
  global: false,
  cascade: true,
  keyframe: false,
  prefix: true,
  compress: false,
  semicolon: false,
});

stylis('html /deep/ .' + componentId, css);
```

TrySound:
Hi @wanbinkimoon. Which bundler do you use?

Andarist:
Closing as no followup information has been given since June.

PR #153, opened 2019-11-21
Better ESM, CJS and UMD compatibility

osdevisnot:
Hi @thysultan, I came across the import issue when trying to use stylis in ESM context. After looking into #81 and #146, I realized that you were working on `v4` branch for next major.

This PR brings in a proposal for `v4` branch to achieve better compatibility for `esm`, `cjs` and `umd` using [klap](https://www.npmjs.com/package/klap) - a tool that I recently started working on to solve this exact problem.

If this PR is merged, we are looking towards a dist folder with below files:
image

I still need to explore the `context: 'this'` part of what you have in rollup config, so please consider this PR as proposal until I figure this out. I wanted to get PR out for discussion TBH.

Also, let me know your thoughts on if we want to have similar compatibility change for master branch instead based on how soon you anticipate v4 to be out.


osdevisnot:
It seems Semaphore does not pick a node version from `.nvmrc` by default. Will have to add Semaphore config for this to pass.

osdevisnot:
It's unlikely for this PR to generate any interest.

PR #155, opened 2019-12-02
Implement POC stringifier and add test setup

Andarist:
This implements a very simple stringifier - as the main complexity lives in `compile` it wasn't actually much to implement here, at least until prefixer, plugins & other stuff get hooked into it.

Additionally, I had to test my work and also wanted to check out how this compares to v3 right now so I've included a test setup. I've chosen jest because I find its watch mode very good and pleasant to work with, especially with typeahead plugin. If you don't like it I can convert those tests to your framework of choice.

The test file contains many comments (many duplicated) - mentioning things I've noticed about broken tests, list of unimplemented stuff and similar. We can move those out of the code to keep the code clean but I find at least those commented out tests quite useful as they can be worked on quicker than if we'd have keep them separately.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "plugins": ["@babel/plugin-transform-modules-commonjs"]
+}
--- a/__tests__/index.js
+++ b/__tests__/index.js
@@ -1,0 +1,1057 @@
+import { compile } from "../src/Compile";
+import { stringify } from "../src/Stringify";
+
+const namespacedStylis = (selector, input) =>
+  stringify(compile(selector ? `${selector}{${input}}` : input));
+const defaultStylis = input => namespacedStylis(".user", input);
+
+// TODO:
+// - `:global()` is unimplemented, but probably right now it's by design - need to confirm, https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/tests/spec.js#L145
+//		- https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/tests/spec.js#L1702
+// - test vendor prefixing:
+//		- https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/tests/spec.js#L553
+//		- https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/tests/spec.js#L700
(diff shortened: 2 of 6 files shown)


.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+  "plugins": ["@babel/plugin-transform-modules-commonjs"]
```
Andarist: this only targets jest right now and allows watch mode to operate on source files, transpiling on the fly

package.json
```diff
@@ -26,11 +26,17 @@
   "module": "index.js",
   "files": [],
   "devDependencies": {
+    "@babel/core": "^7.7.4",
+    "@babel/plugin-transform-modules-commonjs": "^7.7.4",
+    "babel-jest": "^24.9.0",
+    "jest": "^24.9.0",
+    "jest-watch-typeahead": "^0.4.2",
+    "prettier": "^1.19.1",
```
Andarist: At first I've wanted to use jest's inline snapshots for tests but because stylis minifies the output those weren't readable (long lines of minified css).

So basically this stays unused right now, but I would recommend enabling prettier on the codebase 😉 Can do that as part of this PR if you want to.

Andarist:
@thysultan friendly 🏓

__tests__/index.js
```diff
@@ -0,0 +1,1057 @@
...
+// 				backdrop-filter: blur(10px);
+// 			}
+// 		`)).toBe([
+// 			'@supports (display:block){.user{color:red;display:none;}.user h1{color:red;}.user h1 h2{color:blue;}}',
+// 			'@supports (appearance:none){.user{color:red;}}',
+// 			'@supports (backdrop-filter:blur(10px)){.user{backdrop-filter:blur(10px);}}'
+// 		].join(''))
+// })
+
+// TODO: declarations in at-rule are not grouped in a rule
```
Andarist: I've done some digging and compared it to how the same thing gets done without at-rule context. A rule gets added here:
https://github.com/thysultan/stylis.js/blob/c0b8dc3ae17c7ee88710caae29993a10c394aa5f/src/Ruleset.js#L70
but the namespace for `@media` is not  added here, so when declarations in a ruleset are processed they land directly as children of a `@media`.

Because the rule is created here proactively we could probably just create a rule inside of that `@media` proactively as well and it would fix the issue. But the fact the rule gets created proactively makes stylis output empty rules at the moment (it's a possibility when parsing smth like this `namespace{.cls{color:hotpink}}` ) - question is if it's good to create empty rules and just leave it for a serialization to omit them or should we aim for not creating empty rules in the first place?

PR #156, opened 2019-12-10
Apply prefix for margin-inline-*, padding-inline-*

dongseok0:
Added vendor prefix (-webkit-) for margin-inline-start, margin-inline-end, padding-inline-start, padding-inline-end.

I found `margin-inline-*` and `padding-inline-*` are not vendor auto-prefixed by `styled-components` whereas `postcss` does it. And also found that they account on `stylis` for vendor prefixing.

https://caniuse.com/#feat=mdn-css_properties_margin-inline-start
https://caniuse.com/#feat=mdn-css_properties_margin-inline-end

https://caniuse.com/#feat=mdn-css_properties_padding-inline-start
https://caniuse.com/#feat=mdn-css_properties_padding-inline-end

### DIFF
--- a/stylis.js
+++ b/stylis.js
@@ -1250,6 +1250,15 @@
 
 				break
 			}
+			// padding-inline-start, padding-inline-end / margin-inline-start, margin-inline-end
+			case 915:   
+			case 965: { 
+				if (out.indexOf('-inline-') > 5) {
+					cache = out.replace('-inline', '')
+					return webkit + cache + out;
+				}
+				break
+			}
 		}
 
 		return out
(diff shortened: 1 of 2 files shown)


stylis.js
```diff
@@ -1250,6 +1250,22 @@
 
 				break
 			}
+			// margin-inline-start, margin-inline-end: m, a, r
+			case 965: {
```
thysultan: Since the logic within the block is the same for padding and margin:

```diff
+ case 915: case 965: {
```

and test for if indexOf > 5.

Andarist:
I'm afraid that this won't get included in the v3 line. However, it got implemented in the upcoming v4. Hope you understand.

PR #160, opened 2019-12-28
Add support for prefixing margin & padding logical properties

Andarist:


src/Prefixer.js
```diff
@@ -109,6 +109,9 @@ export function prefixer (value, length) {
 			 	(WEBKIT+value+';') +
 			 	(value)
 			)
+		// (margin|padding)-inline-(start|end)
+		case 4095: case 3583: case 4068: case 2532:
+			return (WEBKIT+replace(value, '-inline', '')+';') + (value)
```
Andarist: alternatively we could use regexp like this:
```js
return (replace(value, /(.+)-inline(.+)/, WEBKIT+'$1$2')+';') + (value)
```

not sure what is preferred

thysultan: That works.

Andarist: that as in simple string (committed change) or the regexp one? :p

thysultan: regexp

Andarist: Ok, merging that version then 👍

PR #161, opened 2020-01-06
Fix some linting problems

Andarist:


src/Prefixer.js
```diff
@@ -95,6 +95,8 @@ export function prefix (value, length) {
 				// (inline-)?fl(e)x
 				case 101:
 					return replace(value, /(.+:)([^!]+)(!.+)?/,
+						// TODO: fix this, it's currently broken
```
Andarist: `size` is not declared anywhere, I'm not yet sure how to fix it - what it was supposed to be, seems like some index/pointer to a position in `value` as `+7` is probably an offset to target a hyphen inside `"inline-"`

thysultan: I think it's meant to the position from the colon `:` in `display:inline-flex`

thysultan: Also, i think it was computed because there where certain cases when `"display:"` could be `"display/*inline-comment*/:"` although strange there was that possibility, but now since that's not possible because we don't inline the comments in parsing it shouldn't be a problem to inline size as `8  + 7` where 8 is the length of `"display:"`

Andarist: K, makes sense - whitespace doesn't seem to affect this either, so I've inline `14` (index, not length) which seems to be correct for this.

thysultan: 8 + 7 is 15, does it work with 14?

Andarist: Yes, because we X is supposed to be an index and not a length:
```js
'display:inline-'.charCodeAt(14) === '-'.charCodeAt(0)
```

Andarist: @thysultan I assume this is OK to merge, let me know if you had any other objections

PR #162, opened 2020-01-06
Fix at-rules parsing

Andarist:
This is fixing this test: https://github.com/thysultan/stylis.js/blob/5507a0946e9dc513caac28e3565bff8eeb203283/test/index.js#L319
and partially also this:
https://github.com/thysultan/stylis.js/blob/5507a0946e9dc513caac28e3565bff8eeb203283/test/index.js#L49


test/index.js
```diff
@@ -4,8 +4,6 @@ const stylis = string => serialize(compile(`.user{${string}}`))
 
 describe('PLACEHOLDER', () => {
   // TODO:
-  // - `:global()` is unimplemented, but probably right now it's by design - need to confirm, YES https://github.com/thysultan/stylis.js/blob/4561e9bc830fccf1cb0e9e9838488b4d1d5cebf5/tests/spec.js#L145
```
Andarist: removed "answered" questions from here

src/Parser.js
```diff
@@ -67,7 +67,7 @@ export function parse (value, points, declarations, rules, rulesets) {
 						push(rulesets, ruleset(temporary, points, index, offset, rules, type, props = [], children = []))
 
 						if (character !== 59)
-							if (offset === 0)
+							if (offset === 0 || type === PAGE || type === FONT_FACE)
```
Andarist: I bet you might have some perf trick to make this faster 😅

thysultan: BTW offset is the length of the atrule so: 

```
@supports -> 8
@media -> 5
@font-page -> 9
@page -> 4
@keyframes -> 9
```

and you only need to test for `@media` and `@supports`.

```js
// the above code
else if (parse(value, points, props = [], atrules(offset, rules), children), sizeof(props))
// ... some function
function atrules (offset, rules) {
switch (offset) { case 8: case 5: return [''] }
return rules
}
```

Andarist: According to my tests it's not only about `rules` parameter, but also the last one (`rulesets`). In my working version  local `rulesets` are passed into `parse` but if we go with yours then local `children` are being passed down and this breaks stuff. 

Moreover - in my version first argument is `temporary` which is equal to a `@font-face` (in my test case) wherein your version it is `value` which is equal to `.user`. Not sure which one is correct, but it seems like passing context-like `@font-face` might be desired here.

The tip about testing 8 & 5 is sound, so maybe I should just tweak this if/else with that? Let me know what do you think, maybe what I've investigated here can shed some more light on a more desirable algo for you.

thysultan: What tests fail with this?

```js
else if (parse(value, points, props = [], offset === 6 || offset === 9 ? rules : [''], children) // ...
```

We are going to remove the first parameter being passed to parse because it is not being used at the moment.

Andarist: I assume that I was supposed to also remove the fix I've applied and only use what you have pasted in here. With that both currently failing tests on v4 branch are failing:
```
      at-rules:
      + expected - actual

      -@page{.user{size:A4 landscape;}}@document url(://www.w3.org/),url-prefix(//www.w3.org/),domain(mozilla.org),regexp("https:.*"){.user body{color:red;}}@viewport{.user{min-width:640px;max-width:800px;}}@counter-style list{.user{system:fixed;symbols:url();suffix:" ";}}@-moz-document url-prefix(){.selector{color:lime;}}@page{@bottom-right{{content:counter(pages);margin-right:1cm;}}.user{color:red;width:none;}}
      +@page{size:A4 landscape;}@document url(://www.w3.org/),url-prefix(//www.w3.org/),domain(mozilla.org),regexp("https:.*"){.user body{color:red;}}@viewport{min-width:640px;max-width:800px;}@counter-style list{system:fixed;symbols:url();suffix:" ";}@-moz-document url-prefix(){.user .selector{color:lime;}}@page{color:red;@bottom-right{content:counter(pages);margin-right:1cm;}width:none;}
      
       @font-face:
      + expected - actual

      -@font-face{.user{font-family:Pangolin;src:url('Pangolin-Regular.ttf') format('truetype');}}
      +@font-face{font-family:Pangolin;src:url('Pangolin-Regular.ttf') format('truetype');}
```

Andarist: > We are going to remove the first parameter being passed to parse because it is not being used at the moment.

As far as I can tell it's used by `ruleset` function (`parse` is passing it down there), maybe the usage can be removed though.

thysultan: I see, yes your method works better, though you can replace using TYPE with offset as a comparator. 

I'd turn the whole thing into a switch block and have `@supports` and `@media` and `@document` for one branch and `default` for the others since there are more of the "others" than the former.

The value can be removed as it looks like the reference passed to ruleset is not being used, i.e you can pass any value(i.e temporary) and it will still work.

thysultan: Scratch the last part: i noticed now that ruleset does use `value`, so that's important.

Andarist: Done, tests did not regress in comparison to the previous version of the fix 👍

src/Enum.js
```diff
@@ -10,6 +10,7 @@ export var SUPPORTS = '@supports'
 export var DOCUMENT = '@document'
 export var KEYFRAMES = '@keyframes'
 export var PAGE = '@page'
+export var FONT_FACE = '@font-face'
 export var FONT_FEATURE_VALUES = '@font-feature-values'
```
thysultan:
```diff
-export var FONT_FACE = '@font-face'
```

Andarist: I have actually left this here deliberately. IMHO good to have all possible values listed in a file like this - even if some stay unused in the code (rollup or any other minifier will just remove this anyway).

Let me know if you want to keep this or remove.

thysultan: I don't even think we use any of the at-rule constants there. 

But i see your point, so maybe let's add the other at-rules constants like `@viewport, @counter-style` etc.

src/Parser.js
```diff
@@ -66,11 +66,17 @@ export function parse (value, points, declarations, rules, rulesets) {
 					default:
 						push(rulesets, ruleset(temporary, points, index, offset, rules, type, props = [], children = []))
 
-						if (character !== 59)
-							if (offset === 0)
+						if (character === 59) break
```
thysultan: This is still incomplete, it breaks on other obscure `@at-rules`. I came up with the following

```js
if (character !== 59)
	if (offset === 0)
		parse(temporary, points, children, props, rulesets)
	else if (previous !== 109 && previous !== 115)
		parse(temporary, points, children, [''], children)
	else if (parse(value, points, props = [], rules, children), sizeof(props))
		push(children, ruleset(value, points, 0, 0, rules, type, rules, props))
```

with an amendment to this line(L96): https://github.com/thysultan/stylis.js/blob/8451caba2a67a65ad56a2e1c713a3b8e592ecfa5/src/Parser.js#L96

into the following

```js
character = peek(0), offset = strlen(type = temporary += identifier(caret()))
```

Andarist:
all, currently enabled, tests passing 🎉

src/Enum.js
thysultan: Let's remove at-moz-document, since it's not official. We can remake it with MOZ + DOCUMENT if need me.

Andarist: Ok, I've removed it from this listing - but I still keep the test and support in the code. Or should I remove those as well?

thysultan: Everything else is ok. The code works on the presence of the hyphen, i think this would make it generic for all similar vendor specific at-rules.

Andarist: Ok, merging this in then!

PR #164, opened 2020-01-06
Do not minify "module"

Andarist:


thysultan:
With the stated changes in the comments of the "commit dump" this file will not be in use.

>I'm not convinced that this file should be minified with terser - this is intended to be consumed by bundlers and having it minified might deteriorate DX when debugging during development. Minifying this can be just a consumer's responsibility.

How would that help in "debugging during development" the minified module is sourcemap linked.

Andarist:
Right, I haven't noticed that sourcemaps are being distributed as well. Maybe it's just me - but I actually **hate** sourcemaps because they often trick a reader and it's harder to evaluate expressions in the console (being paused on a breakpoint) because all that can be seen are original identifiers, while you can type & use transpiled ones which in majority of use cases are completely different. Additionally, it's way harder to tinker with the code in node_modules with this approach because you have to delete those sourcemaps (to avoid even more confusion) and if the distributed source code is minified then it's hard to tweak it to test how it behaves. That's why I really prefer readable code in node_modules without any sourcemaps - sourcemaps are (IMHO) somewhat OK for application code where you own the code and can tweak it in your project directly. I don't mind this that much here as a) it's your project so please do whatever you think is best, b) one wanting to debug stylis in node_modules is less likely than for other projects (but not completely unlikely).

thysultan:
Yes, but the module field is used by bundlers who also emit sourcemaps for their sources, though i'm unsure whether this is recursive(it including the sourcemap of the source downstream).

It also allows us to do ahead of time rollup/terser inlining for small functions that can be, and it is also used by `unpkg` when imported as `unpkg/stylis?module`

PR #165, opened 2020-01-08
Enable more tests

Andarist:
This just enables more tests so I'm going to merge this immediately, but doing it as a PR so you can take a look if you want.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -193,7 +193,6 @@ describe('PLACEHOLDER', () => {
   })
 
   // TODO: declarations in at-rule are not grouped in a rule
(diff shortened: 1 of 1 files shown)


test/index.js
```diff
@@ -602,7 +598,7 @@ describe('PLACEHOLDER', () => {
   // })
 
   // TODO: seems broken right now
-  // test('no semi-colons II", () => {
+  // test('no semi-colons II', () => {
   //  expect(stylis(`
```
thysultan: We won't be needing these no semicolon tests.

PR #166, opened 2020-01-08
Fix order of at-rules children

Andarist:


src/Parser.js
```diff
@@ -73,7 +73,7 @@ export function parse (value, points, declarations, rules, rulesets) {
 							else if (attoken(atrule))
 								parse(temporary, points, children, [''], children)
 							else if (parse(value, points, props = [], rules, children), sizeof(props))
-								push(children, ruleset(value, points, 0, 0, rules, type, rules, props))
+								unshift(children, ruleset(value, points, 0, 0, rules, type, rules, props))
```
Andarist: `unshift` is, of course, worse in its perf than `push`, but i don't think there is anything better to do here as we only want to make this a first item in `children` conditionally and we know that it's going to be non-empty after parsing (and pushing to `children`) other stuff first

src/Utility.js
```diff
@@ -93,8 +93,17 @@ export function sizeof (value) {
 /**
  * @param {any[]} value
  * @param {any} element
- * @return {number}
+ * @return {void}
```
Andarist: return value has stayed unused so I thought it might be better to skip returning a value at all to save few bytes from the script

test/index.js
```diff
@@ -192,75 +192,77 @@ describe('PLACEHOLDER', () => {
...
+        @supports (appearance: none) {
+          color:red;
+        }
+        @supports (backdrop-filter: blur(10px)) {
+          backdrop-filter: blur(10px);
+        }
+     `)
+    ).to.equal([
+      '@supports (display:block){.user{color:red;display:none;}.user h1{color:red;}.user h1 h2{color:blue;}}',
+      '@supports (appearance:none){.user{color:red;}}',
```
Andarist: this test still ain't passing completely, because produced output looks like this:
```diff
-@supports (appearance: none)
+@supports (appearance:none)
```
Should we aim to remove this whitespace? If yes - I would prefer working on it in a separate PR though.

thysultan: No need, whitespace in () is treated as special because of whitespace in calc() having semantic meaning.

src/Parser.js
thysultan: It's faster than concat, so i say it's a win. This is only every done once for all the flat declarations in an at-rule nested rule block.

test/index.js
Andarist: right, this situation is slightly different though so we could special case it, not sure what exact goals of stylis are in regards to minification of the output etc

Andarist: anyway - I've added those whitespaces to the expected output and all tests are passing now 🚀

src/Parser.js
Andarist: Ok, this is ready to be merged then, right? 😃

Andarist: Ok, this is ready to be merged then, right? 😃

test/index.js
thysultan: I think this sums up the whitespace issue within calc() functions – https://stackoverflow.com/questions/34419813/why-must-a-or-be-surrounded-with-whitespace-from-within-the-calc-method is feasable to handle it. If you have any idea's you can take a stab, i'll also think about how that could be done, though not a priority right now.

src/Parser.js
thysultan: Yes

test/index.js
Andarist: My main goal is to achieve correctness, so definitely not a priority for me either. I was just asking it as a general question. Thanks for the SO link!

PR #168, opened 2020-01-08
Fix comments being parsed too eagerly

Andarist:


src/Parser.js
```diff
@@ -48,7 +48,14 @@ export function parse (value, points, declarations, rules, rulesets) {
 				break
 			// /
 			case 47:
-				push(rulesets, comment(next()))
+				switch (peek(0)) {
```
Andarist: was thinking about moving this above main `default` case so we could fall through to it for this default case here, but then I couldn't use switch/case because I wouldn't be able to break 2x from this 42 and 47 case, so for now i've decided that it's not worth it

src/Parser.js
```diff
@@ -48,7 +48,14 @@ export function parse (value, points, declarations, rules, rulesets) {
...
+					// *, /
+					case 42: case 47:
+						push(rulesets, comment(next()))
+						break
+					default:
+						temporary += '/'
```
Andarist: the target character is known, so I've just inlined it instead of converting `character` to it

test/index.js
```diff
@@ -264,42 +264,36 @@ describe('PLACEHOLDER', () => {
...
-  // })
+  test('@media specifity', () => {
+    expect(
+      stylis(`
+        > #box-not-working {
+          background:red;
+          padding-left:8px;
+          width:10px;
+          @media only screen and (min-width:10px) {
+            width: calc(10px + 90px * (100vw - 10px) / 90);
```
Andarist: I've decided to move all calcs' here to single lines as this test isn't really concerned about multi-lines values and writing expected result for those is more cumbersome

test/index.js
```diff
@@ -264,42 +264,36 @@ describe('PLACEHOLDER', () => {
...
+          @media only screen and (min-width:10px) {
+            height: calc(10px + 90px * (100vw - 10px) / 90);
+          }
+          @media only screen and (min-width:90px) {
+            height: 90px;
+          }
+         }
+      `)
+    ).to.equal([
+      '.user >#box-not-working{background:red;padding-left:8px;width:10px;height:10px;}',
```
Andarist: the output here has 2 differences from the previous expected results:
1. there is no whitespace after `>`, but I've checked with other CSS parser (prettier) and this seems to be OK, whitespace seems to be optional, so this can be considered a "win" (there is still a whitespace before it though)
2. there is no whitespace around `/`, but this also seems to be optional

src/Parser.js
thysultan: What test does this fix?

Andarist: The one that is enabled in this PR - https://github.com/thysultan/stylis.js/pull/168/files#diff-910eb6f57886ca16c136101fb1699231R275

I've narrowed it down to:
```css
.user{w:2/1;}
```
being parsed to:
```css
.user{w:2;}
```
without this fix. Basically `/` has caused everything after it being treated as a comment.

thysultan: Let's use `font: 12px/14px sans-serif;` as the test for this.

Andarist: ok, added such. Other than that, is it OK to merge this in?

PR #169, opened 2020-01-10
Add tests for escaped quotes

Andarist:


src/Tokenizer.js
```diff
@@ -112,6 +112,8 @@ export function delimit (type) {
 	while (character = scan())
 		if (character === type)
 			break
+		else if (character === 92)
```
thysultan: I pushed a commit that fixes the other delimiter issues, and includes this.

Andarist: Ok, I've adjusted this PR - it only contains new tests now 👍

PR #170, opened 2020-01-10
Add an explicit test for nested parenthesis

Andarist:


Andarist:
@thysultan this still might be worth merging in as an explicit test for this case, WDYT?

thysultan:
Yes.

PR #171, opened 2020-01-10
Add test for & inside a string

Andarist:


thysultan:
I don't see this as an issue. In what real world case will this example reveal itself?

Andarist:
This is, of course, a little bit contrived - but not that unlikely that I would consider this to be a non-issue. Personally, I don't think I would ever run into this, but I can imagine this being a problem for somebody. I can imagine some server-generated localization phrases (or just keys) including `&` put in the data attribute and selectors also generated based on those.

thysultan:
The '&' is an invalid character for a css selector, so that should be a non-issue.

Andarist:
`&` is not used here as part of the selector per se, but only as a part of the attribute value and it seems that this is a legit thing:
https://jsfiddle.net/vacdywzh/

thysultan:
A good real-world-ish test for this could be:

```css
[href="https://css-tricks.com?a=1&b=2"] {...}
```

Andarist:
Ok, I've merged in the proposed test as you have fixed the issue.

PR #173, opened 2020-01-12
Add failing test for & inside a css variable

Andarist:
can't find a compelling real-life use case, but I thought I would bring this up for discussion


thysultan:
>A QA engineer walks into a bar. He orders a beer. Orders 0 beers. Orders 99999999999 beers. Orders -1 beers.

thysultan:
The following regex `/\&(?=([^"']*["'][^"']*["'])*[^"']*$)/g` could probably catch the & as an alternative to the current implementation.

Andarist:
I've prepared a fixing PR - https://github.com/thysultan/stylis.js/pull/174

PR #174, opened 2020-01-14
Fix css variable edge cases

### LINKED ISSUE #144: Stylis can't parse CSS Variables
Hello! I've just tried running known-valid CSS variable values through Stylis and it appears to have quite a few errors with how it's parsing CSS variable values.

Some examples of valid CSS variables that Stylis appears to have problems with include:

```css
--braces: { };
--at-keyword-unknown-block: @foobar {};
--at-keyword-known-block: @media {};
--cdo-at-top-level: <!--;
--cdc-at-top-level: -->;
--semicolon-not-top-level: (;);
--cdo-not-top-level: (<!--);
--cdc-not-top-level: (-->);
```

These values are taken directly from the Web Platform Tests file for what a legal CSS variable value should be: https://github.com/web-platform-tests/wpt/blob/master/css/css-variables/test_variable_legal_values.html#L64&L86

Screen Shot 2019-04-12 at 4 21 53 PM

It's not just these specific examples that are broken, but they seem to highlight that CSS variable values don't appear to be parsed according to the syntax described for them in the [CSS Variables spec](https://www.w3.org/TR/css-variables-1/#syntax) so there are likely many other examples that are valid that would also be broken.

Andarist:
fixes #144 fixes #173

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -27,6 +27,7 @@ export function parse (value, points, declarations, rules, rulesets) {
 	var character = 0
 	var atrule = 0
 	var ampersand = 1
+	var variable = 1
 	var type = ''
 	var props = rules
 	var children = rulesets
@@ -44,22 +45,23 @@ export function parse (value, points, declarations, rules, rulesets) {
 				break
 			// \t \n \s
 			case 9: case 10: case 32:
-				temporary += whitespace(previous)
+				temporary += variable ? whitespace(previous) : from(character)
 				break
 			// /
 			case 47:
 				token(peek(0)) > 2 ? append(comment(scan()), rulesets) : temporary += from(47)
 				break
 			// {
-			case 123:
+			case 123 * variable:
 				points[index++] = strlen(temporary) * ampersand
 			// } ; \0
-			case 125: case 59: case 0:
+			case 125 * variable: case 59: case 0:
 				switch (character) {
 					// \0 }
 					case 0: case 125: --scanning
 					// ;
 					case 59 + offset:
+						variable = 1
 						if (length > 0)
 							append(declaration(temporary, length), declarations)
 						break
@@ -78,10 +80,16 @@ export function parse (value, points, declarations, rules, rulesets) {
 
 				index = length = offset = 0, ampersand = 1, type = temporary = ''
 				break
+			// -
+			case 45:
+				switch (previous) {
+					case 45:
+						variable = 0
+				}
 			default:
 				switch (temporary += from(character), character) {
 					// &
-					case 38:
+					case 38 * variable:
 						ampersand = offset > 0 ? 1 : -1, temporary += from(12)
 						break
 					// :
@@ -89,11 +97,11 @@ export function parse (value, points, declarations, rules, rulesets) {
 						length = strlen(temporary) - 1
 						break
 					// @
-					case 64:
+					case 64 * variable:
 						atrule = peek(0), offset = strlen(type = temporary += identifier(caret())), character++
 						break
 					// ,
-					case 44:
+					case 44 * variable:
 						points[index++] = (strlen(temporary) - 1) * ampersand, ampersand = 1
 						break
 				}
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -26,8 +26,8 @@ export function parse (value, points, declarations, rules, rulesets) {
 	var previous = 0
 	var character = 0
 	var atrule = 0
-	var ampersand = 1
 	var variable = 1
+	var ampersand = 1
 	var type = ''
 	var props = rules
 	var children = rulesets
@@ -61,7 +61,6 @@ export function parse (value, points, declarations, rules, rulesets) {
 					case 0: case 125: --scanning
 					// ;
 					case 59 + offset:
-						variable = 1
 						if (length > 0)
 							append(declaration(temporary, length), declarations)
 						break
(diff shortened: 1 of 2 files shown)


src/Tokenizer.js
```diff
@@ -164,3 +171,38 @@ export function identifier (index) {
 
 	return slice(index, caret())
 }
+
+/**
+ * @return {number}
+ */
+function _variable() {
```
Andarist: those names are ofc not perfect, i couldnt really figure out how to name them based on the naming scheme in stylis, any suggestions?

thysultan:
With the amount of code needed to catch this, i'm starting question, the cost benefit of it.

One possible pseudo algorithm to trim this down and make it palatable from that perspective is the following::

Store a value 1 in a variable called `variable`, when the current character is `-` and the previous character is the same, assign the variable to the number 0. on line 82: `switch (temporary += from(character), character * variable) {` multiple the character with the `variable`, and do the same for  line 57 `case 125 * variable: case 59: case 0:`. 

When ever we enter into a variable context this escapes entering into a `ruleset`, `atrule` or `&` context until we reach a semicolon `;` which then resets the variable back to 1.

I think this might work this the least amount of code.

Andarist:
> With the amount of code needed to catch this, i'm starting question, the cost benefit of it.

Out of curiosity - I've checked and it's 101 bytes (min+gzipped).

> I think this might work with the least amount of code, and best performance.

Gonna play around with this.

Andarist:
Ok, an adjusted version of the proposed solution seems to fix (at least for existing test cases 😅).

src/Parser.js
```diff
@@ -78,22 +80,28 @@ export function parse (value, points, declarations, rules, rulesets) {
...
 				index = length = offset = 0, ampersand = 1, type = temporary = ''
 				break
+			// -
+			case 45:
+				switch (previous) {
+					case 45:
+						variable = 0
+				}
 			default:
 				switch (temporary += from(character), character) {
```
thysultan: why not multiply the character variable once? It is a number.

src/Parser.js
```diff
@@ -78,22 +80,28 @@ export function parse (value, points, declarations, rules, rulesets) {
 
 				index = length = offset = 0, ampersand = 1, type = temporary = ''
 				break
+			// -
+			case 45:
+				switch (previous) {
```
thysultan: `if` should be ok, as it's shorted.

src/Parser.js
```diff
@@ -78,22 +80,28 @@ export function parse (value, points, declarations, rules, rulesets) {
 
 				index = length = offset = 0, ampersand = 1, type = temporary = ''
```
thysultan: the `variable = 1` should go here i.e `variable = ampersand = 1`

src/Parser.js
```diff
@@ -44,22 +45,23 @@ export function parse (value, points, declarations, rules, rulesets) {
 				break
 			// \t \n \s
 			case 9: case 10: case 32:
-				temporary += whitespace(previous)
+				temporary += variable ? whitespace(previous) : from(character)
```
thysultan: we don't need this.

src/Parser.js
Andarist: we want to match `:`, it's important for setting `length`

src/Parser.js
Andarist: k, i thought `switch` is preferable, i thought there might be some perf magic going on in v8 for switch statements, gonna change this tomorrow

src/Parser.js
thysultan: we could move it to the the top switch.

src/Parser.js
Andarist: technically we do, if we want to be 100% correct - when reading css variables values in JS this whitespace matters, while preserving trailing & leading whitespace doesnt really seem important for real-life, im unsure if this would handle OK all the whitespace **in** the value if we remove this, and the whitespace in the value seems important, i can test this tomorrow to see how the whitespace would be handled then, but maybe you know from the top of your head?

src/Parser.js
Andarist: you are most likely right, can you think of any nice test for this?

src/Parser.js
thysultan: In what case would the whitespace be important?

src/Parser.js
thysultan: It's not a semantic change, we just reset everything on one place.

src/Parser.js
Andarist: https://jsfiddle.net/haqxs9ug/

thysultan: >"       white    space   " -> `" white space "`

The browser strips most of the whitespace, how is whitespace important in that fiddle?

Andarist: most - yes, but not all, this changed line makes leading & trailing whitespaces preserved (including others as well, but those will be stripped by browsers anyway - like you have mentioned)

thysultan: Yes, but like CSS i don't see how this is important in CSS. Do you have any case where this has an observable behavior in CSS?

For example consider this:

```css
--rgb: 0, 0, 0;
--rgba rgba(var(rgb), 0.1);
```

Whether it's `0, 0, 0` or `0,0,0` the effect will be the same.

Andarist: nah, cant see any problems with CSS, just when reading with JS, which is also far-fetched

thysultan:
Apart from the whitespace issue, this looks good to me. Let's fix that and merge.

Andarist:
@thysultan ok, just take a look at the whitespace reverting commit and i'm gonna merge this in

thysultan:
LGTM

PR #175, opened 2020-01-14
Remove unnecessary assignment in delimit

Andarist:


src/Tokenizer.js
```diff
@@ -112,7 +112,7 @@ export function dealloc (value) {
  * @return {number}
  */
 export function delimit (type) {
-	while (character = scan())
```
Andarist: `character` is already available in the module scope

thysultan:
LGTM, merge.

PR #176, opened 2020-01-15
Prefixer tests

Andarist:
Please review somewhat carefully - it was a repetitive task and I could make a mistake somewhere

### DIFF
--- a/src/Prefixer.js
+++ b/src/Prefixer.js
@@ -13,11 +13,13 @@ export function prefix (value, length) {
 			return value
 		// animation, animation-(delay|direction|duration|fill-mode|iteration-count|name|play-state|timing-function)
 		case 5737: case 4201: case 3177: case 3433: case 1641: case 4457: case 2921:
-		// text-decoration, filter, mask, clip-path, backface-visibility, column, text-size-adjust
-		case 5572: case 6356: case 6391: case 5844: case 3191: case 6645: case 2756:
+		// text-decoration, filter, mask, clip-path, backface-visibility, column, box-decoration-break
+		case 5572: case 6356: case 6391: case 5844: case 3191: case 6645: case 3005:
+		// columns, column-(count|fill|gap|rule|rule-color|rule-style|rule-width|span|width)
+		case 6389: case 5109: case 5365: case 5621: case 3829:
 			return WEBKIT + value + ';' + value
-		// box-decoration-break, appearance, user-select, flex, transform, hyphens
-		case 3005: case 5349: case 4246: case 6828: case 4810: case 6968:
+		// appearance, user-select, flex, transform, hyphens, text-size-adjust
+		case 5349: case 4246: case 6828: case 4810: case 6968: case 2756:
 			return WEBKIT + value + ';' + MOZ + value + ';' + MS + value + ';' + value
 		// order
 		case 6165:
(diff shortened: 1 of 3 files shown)


test/Prefixer.js
```diff
@@ -3,168 +3,147 @@ import {prefix, compile, serialize} from "../index.js"
...
 	})
 
 	test('clip', () => {
-		expect(prefix(`clip-path: none`)).to.equal([``].join(''))
+		expect(prefix(`clip-path:none`, 9)).to.equal([`-webkit-clip-path:none;`, `clip-path:none`].join(''))
 	})
 
 	test('zoom', () => {
-		expect(prefix(`min-zoom: 0`)).to.equal([``].join(''))
+		expect(prefix(`min-zoom:0`, 8)).to.equal([`min-zoom:0`].join(''))
```
thysultan: Move to line 130-131, i think this is a false positive test for min-inline etc...

test/Prefixer.js
```diff
@@ -3,168 +3,147 @@ import {prefix, compile, serialize} from "../index.js"
...
-		expect(prefix(`backface-visibility: hidden`)).to.equal([``].join(''))
-		expect(prefix(`backface-visibility :hidden`)).to.equal([``].join(''))
+		expect(prefix(`backface-visibility:hidden`, 19)).to.equal([`-webkit-backface-visibility:hidden;`, `backface-visibility:hidden`].join(''))
 	})
 
-	test('transition', () => {
-		expect(prefix(`transition:transform 1s,transform all 400ms,text-transform`)).to.equal([``].join(''))
-		expect(prefix(`transition: transform 1s,transform all 400ms,text-transform`)).to.equal([``].join(''))
-		expect(prefix(`transition: transform 1s,transform all 400ms,text-transform`)).to.equal([``].join(''))
+	test.skip('transition', () => {
```
thysultan: What was the issue here?

src/Prefixer.js
```diff
@@ -90,7 +92,7 @@ export function prefix (value, length) {
 				break
 		// display: (flex|inline-flex|inline-box)
 		case 6444:
-			switch (charat(value, strlen(value) - 2 - (indexof(value, '!important') && 10))) {
+			switch (charat(value, strlen(value) - 2 - (~indexof(value, '!important') && 10))) {
```
thysultan: Does this work, i.e ~10 is not 10 but -11, was this breaking as it where?

test/Prefixer.js
Andarist:
```diff
- actual
+ expected

-transition:-webkit-transform 1s,-webkit-transform all 400ms,text-transform;transition:transform 1s,transform all 400ms,text-transform
+-webkit-transition:text-transform,-webkit-transform 1s,-webkit-transform all 400ms;transition:transform 1s,transform all 400ms,text-transform
```

I didnt have time to dig into this, so I've just skipped this for now - so this can be investigated in a followup PR.

src/Prefixer.js
Andarist: Yes, it wasn't working. We want to subtract 10 if there is `!important` but `-1 /*indexof*/ && 10` is 10 instead of 0, `~` fixes this, it will evaluate to 10 for all return values different than -1

thysultan: Oh, my bad, i misunderstood.

thysultan:
LGTM,.

PR #177, opened 2020-01-23
Plugins draft

Andarist:


test/Plugins.js
```diff
@@ -0,0 +1,16 @@
+import {compile, serialize, pluginMiddleware, prefix} from "../index.js"
+
+describe('prefixer plugin', () => {
+    const prefixPlugin = node => {
+        switch (node.type) {
+            case 'decl':
```
Andarist: i'm not really sure how to prefix correctly `@keyframes` with the current approach used in `prefix` - it would make more sense to me if we could "insert" an additional node, rather than just prepending the prefixed version as a string - the same kinda could be said about prefixing declarations, but it's easier to not do that there (and more performant I bet) so it seems like a reasonable tradeoff there. 

I haven't checked it - but looking at how things are implemented it seems that with the current way of prefixing stuff it only prefixes `@keyframes` itself, but cannot recursively~ prefix its contents.

test/Plugins.js
```diff
@@ -0,0 +1,16 @@
...
+                node.value = prefix(node.value, node.props.length)
+        }
+    }
+
+    const stylis = string => serialize(pluginMiddleware([prefixPlugin])(compile(`.user{${string}}`)))
```
Andarist: this a simplistic version of the API, but I dont know if we can do much better if we want to keep existing `serialize` rather combining plugins & serialize in a single thing (maybe an alternative serialize? `serializeWithPlugins`?)

src/PluginMiddleware.js
```diff
@@ -0,0 +1,32 @@
...
+function run(node, plugins, len) {
+	for (var i = 0; i < len; i++)
+		plugins[i](node)
+}
+
+/**
+ *
+ */
+export function pluginMiddleware(plugins) {
+	return function(ast) {
```
Andarist: because this is a list it prevents us from running a plugin on a whole `ast` right now (without overloading plugins' signature) - maybe worth considering wrapping it in some kind of a root node?

thysultan: Yes, lets have a `root` node.

Andarist:
This got superseded by https://github.com/thysultan/stylis.js/pull/180

PR #178, opened 2020-01-23
Add root node

Andarist:
As requested in https://github.com/thysultan/stylis.js/pull/177#discussion_r370060483

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -92,11 +92,18 @@ export function slice (begin, end) {
 }
 
 /**
+ * @return {number}
+ */
+export function resetcursor() {
+	return line = column = 1, position = 0
+}
+
+/**
  * @param {string} value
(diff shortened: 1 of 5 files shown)


src/Tokenizer.js
```diff
@@ -91,20 +91,27 @@ export function slice (begin, end) {
...
+	return length = strlen(temporary = value), resetcursor()
 }
 
 /**
  * @param {object[]} value
  * @return {object[]}
  */
 export function dealloc (value) {
-	return temporary = '', value
+	return temporary = '', resetcursor(), value
```
Andarist: this was needed to produce correct line mappings to the created root node (from what I understand), as it's created **after** `dealloc` call (because it's the one returning root's children) the cursor is already at the end, but we want to have line mappings pointing to the beginning

PR #180, opened 2020-01-27
middleware proof of concept

thysultan:
Example Use:

```js
serialize(compile(`...`), stringify) // default
serialize(compile(`...`), middleware([prefixer, stringify])) // middlware
function prefixer (current, callback) { if (current.type === DECLARATION) current.value = prefix(current.value) }
```

This tries to avoid a double pass.


src/Serializer.js
```diff
@@ -1,39 +1,74 @@
...
+ */
+export function content (current, prefix, suffix) {
+	return strlen(current) > 0 ? prefix + current + suffix : ''
+}
+
+/**
+ * @param {function[]} current
+ * @return {function}
+ */
+export function middleware (current) {
```
Andarist: It seems to me that `current` is overused here to name multiple different things, it would be great to rename them to something unique because currently, it's somewhat hard to reason about it because of this.

thysultan:
In addition to sourcemap generator, I think that about covers it.

I'll work on the 'cascade: false' and 'global()' ones, it's the reason there's an unused `identifier` function on `./src/Tokenizer.js` that will be a useful utility for plugins along the lines of those two.

src/Tokenizer.js
```diff
@@ -52,16 +69,20 @@ export function token (type) {
 	switch (type) {
 		// * /
 		case 42: case 47:
-			return 3
+			return 5
```
Andarist: could you document those returned numbers anywhere? it's quite hard to grok them without knowing what they mean and how exactly they are supposed to be used

src/Prefixer.js
```diff
@@ -94,21 +92,21 @@ export function prefix (value, length) {
 				break
 		// display: (flex|inline-flex|inline-box)
 		case 6444:
-			switch (charat(value, strlen(value) - 2 - (~indexof(value, '!important') && 10))) {
+			switch (charat(value, strlen(value) - 3 - (~indexof(value, '!important') && 10))) {
```
Andarist: some changes in this file (like this one) look like they are changing the output, but I don't see that many changes in accompanying tests. Do I interpret this wrongly or should I maybe try to prepare appropriate tests?

src/Middleware.js
```diff
@@ -0,0 +1,63 @@
+import {KEYFRAMES, DECLARATION} from './Enum.js'
```
Andarist: If I fix errors in this file (mainly just import missing stuff) all existing tests pass - should I commit it and push out? Or are you in the middle of some work?

src/Middleware.js
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	return stringify(element, callback)
+}
+
+/**
+ * @param {function} callback
+ * @return {function}
+ */
+export function rulesheet (callback) {
```
Andarist: shouldnt this act more like a `middleware`? in the sense that it should accept plugins? `prefixer` shouldnt be included in here by default, right? we should also be able to provide other plugins as well

src/Prefixer.js
Andarist: ok, I think most of the changes are caused by the `;` added to the end of declarations, would love to have a confirmation if that's it or were there some other changes here as well?

src/Parser.js
```diff
@@ -62,9 +58,11 @@ export function parse (value, points, declarations, rules, rulesets) {
 					// ;
 					case 59 + offset:
 						if (length > 0)
-							append(declaration(temporary, length), declarations)
+							append(declaration(temporary + ';', length), declarations)
 						break
-					// rule/at-rule
+					// @ ;
```
Andarist: I'm confused by the `@` on this line, 59 is `;` and I don't know how `@` relates to this branch of code

src/Middleware.js
```diff
@@ -0,0 +1,63 @@
...
+ * @param {string[]} selector
+ * @param {string} namespace
+ * @return {string[]}
+ */
+export function cascade (selector, namespace) {
+	selector.map(function (value) {
+		// h1 h2 :matches(h1 h2) => h1[uuid] h2[uuid] [uuid]:matches(h1 h2)
+		return dealloc(tokenize(alloc(value))).map(function (value, index, children) {
+			switch (charat(value, 0)) {
+				// :global
```
Andarist: It doesnt seem that `:global` handling belongs to this (cascade) plugin. At the high-level they are unrelated and should stay decoupled.

src/Prefixer.js
thysultan: Yes that's it.

src/Middleware.js
thysultan: Middleware.js and Serialize.js are going to change in the next commit.

src/Middleware.js
thysultan: Yes, i've stumbled on an implementation that would allow the previous `middlware([prefixer, stringify, rulesheet])` design to work with this type of plugin.

src/Parser.js
thysultan: It is `;` but ones that a preceded by an at-rule i.e `@import '...';` at-rules terminated with semi-colons.

src/Middleware.js
Andarist: Eager to see what you have come up with!

test/Parser.js
```diff
@@ -1,8 +1,20 @@
-import {compile, serialize} from "../index.js"
+import {compile, serialize, stringify} from "../index.js"
 
-const stylis = string => serialize(compile(`.user{${string}}`))
+const stylis = string => serialize(compile(`.user{${string}}`), stringify)
 
 describe('Parser', () => {
+  test('unested', () => {
```
Andarist: unested? seems like a typo maybe

test/Middleware.js
```diff
@@ -0,0 +1,24 @@
+import {compile, serialize, stringify, middleware, rulesheet, prefixer, namespace} from "../index.js"
+
+const stack = []
+
+describe('Middleware', () => {
+	test('rulesheet', () => {
+  	serialize(compile(`.user{h1 {width:0;} @media{width:1;}}`), middleware([stringify, rulesheet(value => stack.push(value))]))
```
Andarist: the API looks slick 👍

src/Middleware.js
```diff
@@ -0,0 +1,86 @@
...
+ * @param {number} index
+ * @param {object[]} children
+ * @param {function} callback
+ */
+export function prefixer (element, index, children, callback) {
+	if (element.return === element.prefix)
+		switch (element.type) {
+			case DECLARATION: element.prefix = prefix(element.value, element.length)
+				break
+			case KEYFRAMES: element.prefix = prefix(stringify(element, index, element.prefix = null, callback), 10)
```
Andarist: `element.prefix` is a string, is it OK (for engines) to change this property's type to null?

test/Middleware.js
thysultan: 💯 It's 2X faster than current stylis(master branch)

src/Middleware.js
thysultan: Yes, JavaScript engines only special type numbers/doubles.

Andarist:
@thysultan could you prepare a short writeup about the current middleware design? I have started to dig into it but would be a lot easier for me (and future plugin authors) if it would be documented at least a little bit

thysultan:
Thanks

PR #182, opened 2020-02-02
Rulesheet plugin attempt

Andarist:


thysultan:
As an aside, the following seems to log the expected result:

```js
case RULESET: return console.log(element.type, stringify(element, stringify))
case MEDIA: case SUPPORTS: ++context, console.log(element.type, stringify(element, stringify)), --context
```

Andarist:
Any ideas what approach should be explored to handle this?

thysultan:
The following works as a callback middleware:

```js
import {serialize, stringify, prefixer, compile} from 'stylis'

function rulesheet (callback) {
	return function (element, value) { if (value = stringify(element, prefixer)) callback(value) }
}

serialize(compile('...'), rulesheet(value => console.log(value)))
```

Andarist:
Ok, so the idea is to actually replace `middleware` with `rulesheet` - so a `rulesheet` is not a plugin, but rather a "finalizer" (like a `middleware`). I have not thought of that. Going to close this as I've seen you have already included a proposed `rulesheet` in the `middleware` branch.

thysultan:
Yes that was the idea, but i've changed it now, will push my changes later today. This one should be more streamlined.

PR #188, opened 2020-03-09
re-introduce comments into the AST as a node

thysultan:
the changes in src/Tokenizer.js are mostly moving functions around(re-arranging so to speak)

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,14 @@ const declaration = {
 	line: 1, column: 1
 }
 
+const comment = {
+	value: '/*@noflip*/',
+	type: 'comm',
+	props: '/',
+	children: '@noflip',
+	line: 1, column: 1
+}
+
 const ruleset = {
 	value: 'h1,h2',
 	type: 'rule',
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -135,7 +135,7 @@ export function whitespace (type) {
 		else
 			break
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -35,6 +35,14 @@ const declaration = {
 	line: 1, column: 1
 }
 
+const comment = {
+	value: '/*@noflip*/',
+	type: 'comm',
+	props: '/',
```
Andarist: what are possible `props` by design? by running this:
```js
compile(`.user {
  --foo: 10px; // test
}`)
```
I see that inline comment is normalized to a block comment (getting `value: '/* test*/'`) which is fine by me as discussed previously. I'm only wondering if `props: '\n'` is by design for this case.

src/Tokenizer.js
```diff
@@ -19,6 +19,13 @@ export function node (value, root, type, props, children, length) {
 	return {value: value, root: root, type: type, props: props, children: children, line: line, column: column, length: length, prefix: '', return: ''}
 }
 
+/**
+ * @return {number}
+ */
+export function char () {
```
Andarist: maybe `current` would be better for this?

README.md
thysultan: Yes, `/` is for block comments and '\n' for inline comments noted by the last character  used to terminate them.

Andarist: gotcha, cool 👍

PR #189, opened 2020-03-09
Add failing test for "& :hover"

Andarist:


test/Parser.js
```diff
@@ -199,6 +199,16 @@ describe('Parser', () => {
...
   })
 
+  test('& :hover', () => {
+    expect(
+      stylis(`
+        & :hover{
+          color:blue;
+        }
+      `)
+    ).to.equal(`.user :hover{color:blue;}`)
```
Andarist: This is caused by the output of 
[`whitespace`](https://github.com/thysultan/stylis.js/blob/fd51d443f7a0a71db1098e04a4a1b8cd053f76d1/src/Parser.js#L47) being `''`. 

It gets called with `&` and consumes characters till `:` which gives us 0 & 4 tokens respectively which in turn returns us `''` because of some computed token (the one for `:`) being greater than 2.

I've tried adding `&` to isolated tokens list and rewriting [the return statement](https://github.com/thysultan/stylis.js/blob/fd51d443f7a0a71db1098e04a4a1b8cd053f76d1/src/Tokenizer.js#L169) in the `whitespace` but no luck so far (I've managed to pass this test but broke some other ones).

Andarist:
Fixed in https://github.com/thysultan/stylis.js/pull/188 by https://github.com/thysultan/stylis.js/pull/188/commits/a5e6b98cd1b2c037024ce7112a137ac71707a234

PR #190, opened 2020-03-10
Fix "div :hover" case

Andarist:
Followup to https://github.com/thysultan/stylis.js/pull/188/commits/a5e6b98cd1b2c037024ce7112a137ac71707a234

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -45,7 +45,7 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 				break
 			// \t \n \s
 			case 9: case 10: case 32:
-				characters += whitespace(previous, strlen(characters))
+				characters += whitespace(previous)
 				break
 			// /
 			case 47:
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -44,7 +44,7 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 				break
 			// \t \n \s
 			case 9: case 10: case 32:
-				characters += whitespace(previous)
+				characters += whitespace(previous, strlen(characters))
 				break
 			// /
 			case 47:
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -25,6 +25,7 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 	var offset = 0
 	var length = 0
 	var atrule = 0
+	var property = 0
 	var previous = 0
 	var variable = 1
 	var scanning = 1
@@ -61,7 +62,7 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 					// ;
 					case 59 + offset:
 						if (length > 0)
-							append(declaration(characters + ';', rule, length), declarations)
+							append(property > 32 ? declaration(characters + ';', rule, length) : declaration(replace(characters, ' ', '') + ';', rule, length - 1), declarations)
 						break
 					// @ ;
 					case 59: characters += ';'
@@ -87,7 +88,7 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 				break
 			// :
 			case 58:
-				length = strlen(characters)
+				length = strlen(characters), property = previous
 			default:
 				switch (characters += from(character), character * variable) {
 					// &
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -67,24 +67,24 @@ export function slice (begin, end) {
  */
 export function token (type) {
 	switch (type) {
-		// * / comment tokens
+		// * / comment token
 		case 42: case 47:
 			return 6
-		// \0 \t \n \s whitespace tokens
+		// \0 \t \n \s whitespace token
 		case 0: case 9: case 10: case 32:
 			return 5
-		// : accompanied tokens
-		case 58:
-			return 4
-		// ! + , / > @ ~ isolated tokens
+		// ! + , / > @ ~ isolate token
 		case 33: case 43: case 44: case 62: case 64: case 126:
-		// ; { } // breakpoint tokens
+		// ; { } / breakpoint token
 		case 59: case 123: case 125:
+			return 4
+		// : accompanied token
+		case 58:
 			return 3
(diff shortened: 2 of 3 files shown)


src/Parser.js
```diff
@@ -44,7 +44,7 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 				break
 			// \t \n \s
 			case 9: case 10: case 32:
-				characters += whitespace(previous)
+				characters += whitespace(previous, strlen(characters))
```
Andarist: This was introduced to pass this test:
https://github.com/thysultan/stylis.js/blob/31d35e159f833f143eb689c2fe34cb8557cbb088/test/Middleware.js#L19
without it, we were getting double whitespace before `:global()` in things like `width:10px; :global()`. The one whitespace came from the `whitespace` and the second one was received from the `ruleset` when props were joined with whitespace, here:
https://github.com/thysultan/stylis.js/blob/31d35e159f833f143eb689c2fe34cb8557cbb088/src/Parser.js#L135

thysultan: The new test doesn't test this though?

Andarist: I've first introduced this change to the `whitespace`:
```diff
- return token(type) > 2 || token(type === 38 && character === 58 ? type : character) > 2 ? '' : ' '
+ return character !== 58 && (token(type) > 2 || token(character) > 2) ? '' : ' '
```

This has removed the dependency on `&` character - to fix the added test because as shown we need to preserve the whitespace for other selectors followed by pseudo as well. 

This has broken the mentioned `namespace` test because it resulted in double whitespace in the output. Besides the double whitespace (non-ideal output) I couldn't yet find if it actually broke anything semantic-wise, but I've decided to introduce this `!len || ` part to avoid this situation.

thysultan:
The also the case of `color : red;` note the space after `color `.

Andarist:
> The also the case of color : red; note the space after color .

Good catch. So the `strlen(characters)` is not a good criterion to differentiate those 2 cases. My current idea is to find a value that would mean "in the selector context" as in such we want to preserve whitespace preceding a colon and when parsing declarations we don't.

This is quite tricky because we don't quite know this (at least not right now?) when we iterate through the input string. Unless I'm missing something - I have a superficial understanding of the parser as a whole so I very well might miss something. The problem I'm seeing is that we don't know this until we reach `{` or `;`.

Andarist:
@thysultan any ideas on how this could be solved? maybe introducing a needle character like `\f`?

thysultan:
@Andarist The changes(diff) applied to the following functions in src/Tokenizer.js

```js
export function token (type) {
	switch (type) {
		// / comment token
		case 47:
			return 6
		// \0 \t \n \s whitespace token
		case 0: case 9: case 10: case 32:
			return 5
		// ! + , / > @ ~ isolate token
		case 33: case 43: case 44: case 62: case 64: case 126:
		// ; { } / breakpoint token
		case 59: case 123: case 125:
			return 4
		// : accompanied token
		case 58:
			return 3
		// " ' ( [ opening delimit token
		case 34: case 39: case 40: case 91:
			return 2
		// ) ] closing delimit token
		case 41: case 93:
			return 1
	}

	return 0
}
```

```js
export function tokenizer (children) {
	while (next())
		switch (token(character)) {
			case 3: case 0: append(identifier(position - 1), children)
				break
			case 2: append(delimit(character), children)
				break
			default: append(from(character), children)
		}

	return children
}
```

and in `src/Parser.js`

```js
// ;
case 59 + offset:
	if (length > 0)
		append(property > 32 ? declaration(characters + ';', rule, length) : declaration(replace(characters, ' ', '') + ';', rule, length - 1), declarations)
	break
```

Would get rid of the extra space only when that edge case is encounter.

You might need to adjust some expected test output for whitespace after those changes.

Andarist:
👍 so my intuition that this cannot be fixed while parsing was good 😅

I've pushed out changes. A couple of remarks:
1. this builds upon my previous, incomplete, fix
2. I had to bring back `42` (`*`) as a comment token to the `token` function
3. I didn't have to adjust any existing tests, just added a new one for `color : hotpink;`

test/Parser.js
```diff
@@ -486,8 +486,14 @@ describe('Parser', () => {
         div {
           ${'width:0;    '}
         }
+        .foo {
```
thysultan: Are we missing the test for `div :hover`?

thysultan:
>I had to bring back 42 (*) as a comment token to the token function

Why? the following should suffice.

```js
return token(type) > 2 || token(character) > 3 ? '' : ' '
```

I forgot to include it in the comments before this.

test/Parser.js
Andarist: no, it's here: https://github.com/thysultan/stylis.js/pull/190/files#diff-2c21dd491e96fcd23ee0c155c2141f07R212

but i can move it from a separate test to this "whitespace" test, if we do this then the second similar test for `& :hover` should be moved as well, although it's now basically the same situation as `div :hover`

Andarist:
> I forgot to include it in the comments before this.

Added and it's working 👌

Andarist:
Actually - I was only able to remove my previous fix with passing the length to the `whitespace` etc.

I can't remove `*` from being treated as comment token because this breaks:
https://github.com/thysultan/stylis.js/blob/1f6857180e664c31abb453ce49cb103b3e5c79ed/src/Parser.js#L52

src/Tokenizer.js
```diff
@@ -67,24 +67,24 @@ export function slice (begin, end) {
  */
 export function token (type) {
 	switch (type) {
-		// * / comment tokens
+		// * / comment token
 		case 42: case 47:
```
thysultan: What test fails when you remove the `*(42)`? andd is it just a whitespace diff?

Andarist: basically each test containing a block comment, because Parser peeks next character when it meets `/`, it doesnt know if it should start consuming a comment or not unless it makes sure that the next character is `/` or `*`, so removing `*` from here makes Parser not consuming block comments at all

PR #191, opened 2020-03-17
Add a failing test case for rulesheet plugin in combination with prefixed keyframes

Andarist:
Problem is that the return property contains both prefixed and original keyframes in a single string after being processed by `stringify`:
```js
{
    value: '@keyframes foo',
    root: null,
    type: '@keyframes',
    props: [ 'foo' ],
    children: [ [Object], [Object] ],
    line: 1,
    column: 54,
    length: 0,
    prefix: '@-webkit-keyframes foo{from{width:0;}to{width:1;}}',
    return: '@-webkit-keyframes foo{from{width:0;}to{width:1;}}@keyframes foo{from{width:0;}to{width:1;}}'
}
```
and this gets passed to the `rulesheet` plugin as a single rule which is incorrect.

### DIFF
--- a/test/Middleware.js
+++ b/test/Middleware.js
@@ -4,8 +4,13 @@ const stack = []
 
 describe('Middleware', () => {
 	test('rulesheet', () => {
-  	serialize(compile(`.user{h1 {width:0;} @media{width:1;}}`), middleware([stringify, rulesheet(value => stack.push(value))]))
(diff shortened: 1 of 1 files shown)


Andarist:
This has been fixed by https://github.com/thysultan/stylis.js/pull/195/commits/666f4b674a922884580fe3a61e93c2d965b58a4a

PR #192, opened 2020-03-18
Fixed double declarations caused by the prefixer

Andarist:
Currently:
```js
const stylis = styles => serialize(compile(styles), middleware([prefixer, stringify]))

stylis(`body{background:white;}`) === 'body{background:white;background:white;}'
```
After the fix
```js
stylis(`body{background:white;}`) === 'body{background:white;}'
```

I've looked through all the other prefixing rules and it seems to me that those were the only affected ones (other are already guarded somehow or should always produce a prefixed value), but you can double-check this.

### DIFF
--- a/src/Prefixer.js
+++ b/src/Prefixer.js
@@ -7,6 +7,8 @@ import {hash, match, charat, strlen, indexof, replace} from './Utility.js'
  * @return {string}
  */
 export function prefix (value, length) {
+	var input = value
+
 	if (length > 0)
 		switch (hash(value, length)) {
 			// @keyframes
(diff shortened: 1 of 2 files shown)


src/Prefixer.js
```diff
@@ -53,7 +55,9 @@ export function prefix (value, length) {
 				value = replace(value, /(.*)(zoom-\w+|grab\w*)(.*)/, '$1' + WEBKIT + '$2$3$1' + MOZ + '$2$3')
 			// background, background-image
 			case 5495: case 3959:
-				return replace(value, /([^-])(image-set\()/, '$1' + WEBKIT + '$2')
```
thysultan: Why not regex test for image-set then execute the replace i.e

```js
return test(value, /image-set/) ? replace(...) : value
```

Andarist: this was my first attempt but then the cursor has stopped working because it falls through here, so to handle `cursor:pointer;` case we'd probably have to introduce 2 regexp tests and a temporary variable

Andarist:
This has been fixed in https://github.com/thysultan/stylis.js/pull/195/commits/666f4b674a922884580fe3a61e93c2d965b58a4a

PR #193, opened 2020-03-18
Add flex-direction prefixing

Andarist:


src/Prefixer.js
```diff
@@ -24,6 +24,9 @@ export function prefix (value, length) {
 			// appearance, user-select, flex, transform, hyphens, text-size-adjust
 			case 5349: case 4246: case 6828: case 4810: case 6968: case 2756:
 				return WEBKIT + value + MOZ + value + MS + value
+			// flex-direction
```
Andarist: in v3 "flex" was also in this group, so probably it would be safe to move it here to achieve a shorter output, WDYT?

thysultan: What do you mean? "flex" by itself?

Andarist: Exactly, in v3:
```js
stylis('.user', 'flex:1;') === `.user{-webkit-flex:1;-ms-flex:1;flex:1;}`
```
but in v4 currently the equivalent produces `.user{-webkit-flex:1;-moz-flex:1;-ms-flex:1;flex:1;}`

thysultan: Added the Patch upstream(v4 branch) with a lace of other fixes(all current PR's save one).

Andarist: Closing this PR as it got fixed in the mentioned https://github.com/thysultan/stylis.js/pull/195/commits/666f4b674a922884580fe3a61e93c2d965b58a4a

PR #194, opened 2020-03-18
Universal selector bug

Andarist:


src/Tokenizer.js
```diff
@@ -135,7 +135,7 @@ export function whitespace (type) {
 		else
 			break
 
-	return token(type) > 2 || token(character) > 3 ? '' : ' '
+	return token(type) > 2 || character !== 42 && token(character) > 3 ? '' : ' '
```
Andarist: I have noticed now that this still doesnt produce a correct output for `* *` - which is a bizarre thing to write, but it's semantically correct, need to rethink how this can be solved succinctly

Andarist: WDYT about this:
```diff
diff --git a/src/Tokenizer.js b/src/Tokenizer.js
index 8935559..4ad060c 100644
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -67,8 +67,10 @@ export function slice (begin, end) {
  */
 export function token (type) {
        switch (type) {
+               case 42:
+                       return 7
                // * / comment token
-               case 42: case 47:
+               case 47:
                        return 6
                // \0 \t \n \s whitespace token
                case 0: case 9: case 10: case 32:
@@ -124,6 +126,15 @@ export function tokenize (value) {
        return dealloc(tokenizer(alloc(value)))
 }
 
+function helper(t, lower) {
+       return (t = token(t)) > lower && t < 7
+}
+
 /**
  * @param {number} type
  * @return {string}
@@ -135,7 +146,7 @@ export function whitespace (type) {
                else
                        break
 
-       return token(type) > 2 || character !== 42 && token(character) > 3 ? '' : ' '
+       return helper(type, 2) || helper(character, 3) ? '' : ' '
 }
 
 /**
```

Andarist: @thysultan any thoughts on this?

thysultan: We could also divorce the `Parse.js/parse#51` line from the token function entirely.

```js
// /
case 47:
	switch (peek()) {
		// * /
		case 42: case 47:
			append(comment(commenter(next(), caret()), root), declarations)
			break
		default:
			characters += '/'
	}
```

the gzip savings are inconsequential.

Andarist: Adjusted as suggested

Andarist:
This has been fixed in https://github.com/thysultan/stylis.js/pull/195/commits/666f4b674a922884580fe3a61e93c2d965b58a4a

PR #195, opened 2020-03-25
V4

thysultan:
It is just about close to being ready.


src/Utility.js
```diff
@@ -42,7 +48,7 @@ export function test (value, pattern) {
  * @return {string?}
  */
 export function match (value, pattern) {
-	return pattern.exec(value)[0]
+	return test(value, pattern) ? pattern.exec(value)[0] : null
```
Andarist: im curious - are those optimized by engines so subsequent calls on the same inputs are cached? or is `exec` more heavy? any particular reason why this has been chosen instead of:
```js
return (value = pattern.exec(value)) ? value[0] : value
```

thysultan: That's probably better.

Andarist: My proposed version?

thysultan: Yes.

Andarist: K, I've pushed the change directly to this branch

src/Middleware.js
```diff
@@ -0,0 +1,104 @@
...
+ * @param {number} index
+ * @param {object[]} children
+ * @param {function} callback
+ */
+export function prefixer (element, index, children, callback) {
+	switch (element.type) {
+		case DECLARATION: element.return = prefix(element.value, element.length)
+			break
+		case KEYFRAMES:
+			return serialize([assign({}, element, {type: '', value: replace(element.value, '@', '@' + WEBKIT)})], callback)
```
Andarist: 1. shouldnt those "copies" receive a meaningful `.type`? I would expect the "inserted" node to be resembling one that would be created when compiling
2. isn't usage of `assign` here creating a different hidden class than what is returned from `node` and thus making functions receiving this polymorphic which we should avoid?

thysultan: Point.2 is solved with the most recent commit, 

Point 1, that was intentional. The intention is to be able to distinguish between plugin generated rulesets and others, this signals that it is a no-op/completed node as far as a plugin is concerned so other plugins don't mistaken it for a source node and try to process it in their logic, it may make sense for other plugins to preserve the `.type` but in this is i thought it best not.

kripod:
Awesome work, I highly appreciate all the efforts towards the next release! I would like to use v4 for [glaze](https://github.com/kripod/glaze), a CSS-in-JS library in the works. How may I apply autoprefixing on object styles optimally? For instance, I would like to prefix an object similar to:

```js
const style = {
  display: 'flex',
  justifyContent: 'space-evenly',
};
```

Also, is there a way to prefix key-value pairs one by one? If so, how is that possible?

thysultan:
>is there a way to prefix key-value pairs one by one

The `Prefixer.js` tests show how this is done. That would be the best way.

Andarist:
@kripod keep also in mind that stylis operate on a string input, not on objects - so to handle this properly you need to give it a serialized input or incorporate Prefixer into your serialization process

kripod:
Thank you for the feedback! Is there a chance for publishing an early preview version on npm?

package.json
```diff
@@ -8,9 +8,12 @@
   "repository": "https://github.com/thysultan/stylis.js",
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "type": "module",
-  "main": "dist/stylis.umd.js",
+  "main": "dist/stylis.cjs",
```
thysultan, line 11: This should be 'umd.js' unless i've missed something?

Andarist: I've merged the other PR after seeing [the LGTM comment](https://github.com/thysultan/stylis.js/pull/207#issuecomment-614642816). Changed this to UMD just now: https://github.com/thysultan/stylis.js/pull/195/commits/395454b8e75c627db52dc5cd72b91e169c18d774

kripod:
Congratulations on the big merge! 🎉

I cannot wait to try integrating v4 into [glaze](https://github.com/kripod/glaze), primarily as a runtime autoprefixer.

danieldelcore:
Well done everyone! The hard work is appreciated ❤️

PR #202, opened 2020-04-09
Update README.md

danieldelcore:
Minor readme tweek  :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ var stylis = new stylis(options)
 var stylis = stylis
 ```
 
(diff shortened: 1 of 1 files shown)


Andarist:
Thanks for doing this! I'm not sure if it's worth merging this right now as v3 won't receive any updates and we are very close to shipping v4, there is an open PR for this - https://github.com/thysultan/stylis.js/pull/195 , so merging this one would probably only cause conflicts with the that one.

danieldelcore:
Oh sorry i had no idea. I'll close this for now 😃 
Looking forward to v4 🎉

PR #203, opened 2020-04-10
Fix issues around prefixing pseudos

Andarist:
When working on `.parent` I've noticed that `ruleset` did not receive its final argument (`length`) in some cases which resulted in a "broken node" (`node.length === undefined`). After some digging into what `.length` actually does on rule nodes I was able to come up with a failing test for the code before the fix

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -16,12 +16,12 @@ export function compile (value) {
  * @param {string[]} rule
  * @param {string[]} rules
  * @param {string[]} rulesets
- * @param {number} pseudo
+ * @param {number} inpseudo
  * @param {number[]} points
  * @param {string[]} declarations
  * @return {object}
  */
-export function parse (value, root, rule, rules, rulesets, pseudo, points, declarations) {
+export function parse (value, root, rule, rules, rulesets, inpseudo, points, declarations) {
 	var index = 0
 	var offset = 0
 	var length = 0
@@ -32,6 +32,7 @@ export function parse (value, root, rule, rules, rulesets, pseudo, points, decla
 	var scanning = 1
 	var ampersand = 1
 	var character = 0
+	var pseudo = 0
 	var type = ''
 	var props = rules
 	var children = rulesets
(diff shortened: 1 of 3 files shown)


thysultan:
Compare the diff of these two files with your own counter-parts: https://gist.github.com/thysultan/186b10150fce56d40f7cac2b165780c8

Should get an idea of how to fix this.

Andarist:
@thysultan well, my fix also works - so far, do u know how my approach fails? when both yield different results?

thysultan:
```js
a::placeholder{x:0;b{y:0;}}
```

The current fix is specific to only nested at-rules.

Andarist:
I've fixed the issue you have spotted. Also adjusted match for pseudo similarly to what you have had in the gist and added a test for this case, while doing this I've also noticed that pseudo prefixer was replacing false positives so I've fixed it as well. Please also take notice that I've adjusted regexps around pseudo prefixing slightly so they match between match and replace steps.

Andarist:
I've pushed out this proposed fix for the orphaned `:read-write`. Couldn't figure out an alternative solution to this. The nice trait of this fix is that rules' `.length` is always 0 or 1 (so a flag) rather than any, kinda random, integer.

src/Parser.js
```diff
@@ -74,27 +76,27 @@ export function parse (value, root, rule, rules, rulesets, points, declarations)
 					case 59: characters += ';'
 					// { rule/at-rule
 					default:
-						append(reference = ruleset(characters, root, index, offset, rules, points, type, props = [], children = [], length), rulesets)
+						append(reference = ruleset(characters, root, index, offset, rules, points, type, props = [], children = [], (pseudo = pseudo || inpseudo)), rulesets)
```
thysultan, line 79: Can you try this diff: https://gist.github.com/thysultan/186b10150fce56d40f7cac2b165780c8#file-parser-js (updated)

Andarist: it doesn't work for [the orphaned `:read-write` case](https://github.com/thysultan/stylis.js/blob/c4159ecb3a88af498d9a2128e0306f81756c1da1/test/Middleware.js#L77-L80). The problem here is that `:` comes first and thus `length` (used as `pseudo`) is not initialized to a positive integer - it's always 0 for a case like this - and that makes it hard to reuse `length` as `pseudo`.

I've also checked size of both approaches - ur has at the moment 9 bytes less than mine

thysultan: can't we use the property variable?

thysultan: for example:

Parser.js#72

```diff
- append(property > 32 ? declaration(characters + ';', rule, length - 1) : declaration(replace(characters, ' ', '') + ';', rule, length - 2), declarations)
+ append(property > 32 ? declaration(characters + ';', rule, length - 1) : declaration(replace(characters, ' ', '') + ';', rule, length - 2), declarations)
```

Parser.js#98

```diff
- length = strlen(characters), property = previous
+ length = 1 + strlen(characters), property = previous
```

Andarist: I've applied your fix - although overall I think mine was more clear (if it has worked for all cases) and it had this cool effect of `.length` being 0 or 1 for rules. Stylis is all about squeezing the most we can and your version is better by 8 bytes, so I'm fine with that 😉

thysultan: LGTM

PR #204, opened 2020-04-12
Add parent property to nodes

Andarist:


src/Tokenizer.js
```diff
@@ -25,7 +25,7 @@ export function node (value, root, type, props, children, length) {
  * @param {string} type
  */
 export function copy (value, root, type) {
-	return node(value, root.root, type, root.props, root.children, 0)
+	return node(value, root.root, type, root.props, null, root.children, 0)
```
thysultan, line 28: Any reason we aren't carrying the root.parent reference over?

src/Parser.js
```diff
@@ -157,7 +157,7 @@ export function ruleset (value, root, index, offset, rules, points, type, props,
  * @return {object}
  */
 export function comment (value, root) {
-	return node(value, root, COMMENT, from(char()), substr(value, 2, -2), 0)
+	return node(value, root, COMMENT, from(char()), root, substr(value, 2, -2), 0)
```
thysultan, line 160: Why are we re-using the root reference, i'd assume we'd need to pass a new parameter to these functions: `parent`?

src/Parser.js
```diff
@@ -75,7 +75,7 @@ export function parse (value, root, rule, rules, rulesets, pseudo, points, decla
 					case 59: characters += ';'
 					// { rule/at-rule
 					default:
-						append(reference = ruleset(characters, root, index, offset, rules, points, type, props = [], children = [], length), rulesets)
+						append(reference = ruleset(characters, root, index, offset, rules, points, type, props = [], reference, children = [], length), rulesets)
```
thysultan, line 78: Reference is not a stable reference to `rule`.

src/Parser.js
Andarist: As far as I can tell - comment and declaration node points to their parent from the `.root` property, so this is "only" providing consistency with rule nodes (so each get non-null `.parent` when appropriate)

src/Tokenizer.js
Andarist: I was thinking about 3 possibilities here:
1. carrying over root.parent (actually my initial choice)
2. null
3. pointing to the copied node - thought about it briefly, seemed cute at first but hard to tell apart the copied node from the original so I've discarded this quickly

In the end, I've chosen null - to keep `.parent` only for "real" nodes. Kinda similarly to how they always get `.length === 0` (regardless of the original value). I definitely see the value of carrying `.parent` over, so I'm totally OK with changing this.

src/Parser.js
Andarist: Right, good catch - it should be `rule` here. Usage of `reference` for the latter call seems OK though, could you confirm that?

Andarist:
@thysultan I've adjusted the code to your suggestions, please take a second look

src/Parser.js
thysultan: oh, i missed the fact that they where the `comment/declaration` functions.

src/Tokenizer.js
```diff
@@ -12,11 +12,12 @@ export var characters = ''
...
  * @param {string} type
  * @param {string[]} props
+ * @param {object|null} parent
  * @param {object[]} children
  * @param {number} length
  */
-export function node (value, root, type, props, children, length) {
-	return {value: value, root: root, type: type, props: props, children: children, line: line, column: column, length: length, return: ''}
+export function node (value, root, type, props, parent, children, length) {
+	return {value: value, root: root, type: type, props: props, parent: parent, children: children, line: line, column: column, length: length, return: ''}
```
thysultan, line 20: I'm sentimental about `type`, `props`, `children` always being one after another.

src/Parser.js
Andarist: When thinking about this more I thought I would have to adjust the code for those functions to accept `parent` after all as in:
```
compile(`a{@media{w:1;}}`)
```
the parent of this declaration should be `@media`. So I went back checking things and it turns out that the `.parent` is `@media` for this - because it's also the `.root`, which in turn leads me to a question - should the `.root` of this declaration be `@media` or `.children[0]` of this `@media` node? I thought it's the latter, but maybe I'm just misunderstanding the exact purpose of `.root`?

src/Tokenizer.js
Andarist: Adjusted all function to always accept `parent` after the `root`

src/Parser.js
thysultan: It is for ruleset nodes, the `.root` for `decl`/`comm` is as it should be, i.e it will also be `@viewport` for `@viewport{width:device-width;}`.

thysultan:
LGTM, have you experimented with that emotion plugin with theses changes to see if it affords the leverage you'd need to implement it?

Andarist:
> LGTM, have you experimented with that emotion plugin with theses changes to see if it affords the leverage you'd need to implement it?

Yes, existing tests pass - but I need to add more, I already see this implementation here is broken for a couple of cases, for example:
```js
compile(`.user{@media{@supports{b:1}}}`)
```
here the declaration should point to `@supports` and `@supports` should point to `@media`, which is not the case. Need to dig deeper into the algorithm to fix this - gonna ping u back once I figure this out.

src/Parser.js
Andarist: To what declaration's `.root` should point here?
```js
compile(`.user{@media{@supports{b:1}}}`)
```

thysultan:
>here the declaration should point to @supports and @supports should point to @media, which is not the case.

What do they point to as it stands?

Andarist:
I think I got it right this time. However, `parse` has 9 parameters now - is that OK? Not sure if engines treat all functions equally (in that regard). It has somewhat weird effect that for this:
```
.user{@media{w:1;}}
```
we end up with a `@media {.user{}}` rule pointing to `@media` as its parent while having `w:1;` declaration as to its child which in turn **also** points to `@media` as its parent. While a little bit weird at first, it is semantically correct - it would be weirded if that rule (which is always automatically created even if there are no declarations for it) would have been pointing to anything else.

> > here the declaration should point to @supports and @supports should point to @media, which is not the case.

> What do they point to as it stands?

that declaration was pointing to auto-created `@supports{.user{}}` and `@supports` was pointing to auto-created `@media{.user{}}`

Andarist:
@thysultan changed to use that syntax

src/Parser.js
```diff
@@ -13,7 +13,7 @@ export function compile (value) {
 /**
  * @param {string} value
  * @param {object} root
- * @param {object|null} parent
+ * @param {?object} parent
```
thysultan, line 16: I've been using the {type?} suffix placement style.

Andarist: K, adjusted it to the proposed syntax

thysultan:
LGTM

PR #205, opened 2020-04-13
Remove unnecessary assignment

Andarist:
This value is only used inside this block:
https://github.com/thysultan/stylis.js/blob/24f984a77b2ccb41c5a2bf273b9af4be09c602c1/src/Parser.js#L78-L91
but it is actually reassigned at its very beginning, so the value with which it gets initialized is never used and the assignment can be removed


thysultan:
it's assigned to rule because it's later assigned to a rule node with the same object type/map.

Andarist:
Makes sense, was suspecting this might be something like that but wasn't sure if this matters for local variables.

PR #206, opened 2020-04-13
Remove char getter - use character as live binding instead

Andarist:
Not sure if this is preferable but exported values act as live bindings so we don't actually need an extra getter to access their current value.


thysultan:
let's leave it, minifiers and javascript engines optimize these type of functions away.

Andarist:
Actually, Terser is not able to inline this - even if I remove both `character` and `char` from being exported from the final bundle.
```js
// char
function J() {
  return F;
}

// comment
function ce(e, r) {
  return H(e, r, c, d(J()), C(e, 2, -2), 0);
}
```

I'm cool with leaving this as is - was just an idea I had and wanted to see what you are thinking about this.

PR #207, opened 2020-04-14
Add CJS export to conditional exports

Andarist:
Without this, it wouldn't be possible to `require('stylis')` in nodes supporting conditional exports as `require` cannot load modules (in contrary to `import` which can load commonjs scripts)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,9 +8,12 @@
   "repository": "https://github.com/thysultan/stylis.js",
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "type": "module",
-  "main": "dist/stylis.umd.js",
+  "main": "dist/stylis.cjs",
   "module": "dist/stylis.esm.js",
-  "exports": "./index.js",
+  "exports": {
+    "import": "./index.js",
+    "require": "./dist/stylis.cjs"
+  },
   "files": [
     "index.js",
     "dist/",
--- a/script/build.js
+++ b/script/build.js
@@ -1,8 +1,8 @@
 import {join} from 'path'
-import {terser} from "rollup-plugin-terser"
+import {terser} from 'rollup-plugin-terser'
 import size from 'rollup-plugin-size'
 
-const options = {mangle: true, compress: false}
+const options = {mangle: true, compress: false, toplevel: true}
 const defaults = {
 	onwarn(warning, warn) {
 		switch (warning.code) {
@@ -18,6 +18,12 @@ const defaults = {
 
 export default ({configSrc = './', configInput = join(configSrc, 'index.js')}) => {
 	return [
+		{
+			...defaults,
+			input: configInput,
+			output: [{file: join(configSrc, 'dist', 'stylis.cjs'), format: 'cjs', name: 'stylis', freeze: false, sourcemap: true}],
+			plugins: [terser(options), size()]
+		},
 		{
 			...defaults,
 			input: configInput,


package.json
```diff
@@ -8,9 +8,12 @@
   "repository": "https://github.com/thysultan/stylis.js",
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "type": "module",
-  "main": "dist/stylis.umd.js",
+  "main": "dist/umd/stylis.umd.js",
   "module": "dist/stylis.esm.js",
-  "exports": "./index.js",
+  "exports": {
+    "import": "./index.js",
```
Andarist, line 14: I personally don't quite like that a different file is being given to node (pkg.exports.import) and bundlers (pkg.module) - if the code is not specific to its environment.

Could we just point to the same file from both? I'm still not a fan of giving bundlers pre-minified code - had to debug few 3rd part dependencies in the past weeks and this would be a total pain if they would be minified (luckily, they weren't).

script/build.js
```diff
@@ -1,7 +1,16 @@
 import {join} from 'path'
-import {terser} from "rollup-plugin-terser"
+import {terser} from 'rollup-plugin-terser'
 import size from 'rollup-plugin-size'
 
+const emitPackageScopeFile = (type) => {
```
Andarist, line 5: this is required so that commonjs/umd file can be treated as commonjs script while having `.js` extension (otherwise it gets treated as a module due to the closest package.json - the root one - having `"type":"module"`)

package.json
```diff
@@ -8,9 +8,12 @@
   "repository": "https://github.com/thysultan/stylis.js",
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "type": "module",
-  "main": "dist/stylis.umd.js",
+  "main": "dist/umd/stylis.umd.js",
```
TrySound, line 11: Shouldn't commonjs modules have .cjs extension? I think this is a requirement when "type" is "module". The way around is package.json with "{ type: "commonjs" }" in subfolder.

Andarist: @TrySound - see a comment [here](https://github.com/thysultan/stylis.js/pull/207#discussion_r407931392) 😉

TrySound: Nice!

thysultan: Why the new directory?

TrySound: To add another package.json with { type: commonjs }

thysultan:
>it wouldn't be possible to require('stylis')

Examples? As far as i can tell it should work without this as node versions that don't support modules will use the `pkg.main` field. It's unclear why we need this considering that.

Andarist:
It's because in ESM-supporting node (v13) only `import` loader is capable of loading modules. This is documented [here](https://nodejs.org/api/esm.html#esm_conditional_exports):
> "require" - matched when the package is loaded via require(). As require() only supports CommonJS, the referenced file must be CommonJS.

I have also tested this myself before preparing this PR - can't require module, but can import a script (cjs) 🤷

thysultan:
>only import loader is capable of loading modules

Yes that's the intended idea.

Andarist:
Problem is that it means that **all dependents in the chain** must be modules and I'm afraid that the ecosystem is just not ready yet for this. In case of Emotion - nor React, nor Emotion distribute conditional exports just yet so while SSRing node will reach for their respective CJS files and try to access stylis on which it will fail because it's just a module without cjs fallback.

thysultan:
Primary dependents should be bundling stylis and shipping that instead of including it as a direct dependency. This is just for people using stylis directly with "module" supported versions of NodeJS.

Andarist:
Do you say that we at Emotion should bundle stylis? I wouldn't have expected that 😅 i much prefer not doing that as we would need to prepare a release after each new stylis release, but if that's your stance - sure, we can do that. What are the benefits of this approach though? Should this information be put in the docs?

An additional thing I don't quite like about this is that having stylis as a direct dependency would work when using node <=12 but suddenly break when upgrading to a newer one. From the consumer point of view, this is super unexpected.

package.json
```diff
@@ -8,9 +8,12 @@
...
   "module": "dist/stylis.esm.js",
-  "exports": "./index.js",
+  "exports": {
+    "import": "./index.js",
+    "require": "./dist/umd/stylis.umd.js"
```
thysultan, line 15: Why can't this point directly to the file?

thysultan: By directly i mean directly to the current file `/dist/stylis.umd.js`

thysultan: For example in https://nodejs.org/api/esm.html#esm_approach_2_isolate_state they don't try to isolate the cjs file into a directory.

Andarist: > Why can't this point directly to the file? Doesn't the require imply cjs.

Unfortunately not, this is also kinda quirky for me and counter-intuitive. `"require"`/`"import"` keys only tell which loader should be used, but import is able to load cjs files, so it can't set the mode of the interpretation of the imported file. Maybe it's just a symmetry thing for `"require"`?

> For example in https://nodejs.org/api/esm.html#esm_approach_2_isolate_state they don't try to isolate the cjs file into a directory.

Yes, but they use `.cjs` extension - which sets the mode of the interpretation. `.mjs` files are always modules, `.cjs` are always scripts. I've just wanted to avoid `.cjs` extension here, especially given the fact that this is a UMD bundle which often is used when trying to use a particular lib directly in a browser, and `.cjs` doesn't make much sense in such context. We can, however, use `.cjs` to avoid this extra directory.

To sum it up:
- "require"/"import" - don't do anything in regards to module/script interpretation
- `.mjs`/`.cjs` extensions force a particular mode
- `.js` are parsed based on the nearest package.json and its "type", that's why we can create a "nested package scope", as the algorithm allows this, even though this section doesn't really mention this explicitly: https://nodejs.org/api/esm.html#esm_package_scope_and_file_extensions . This technique is used by some packages so a `.js` extension can be used for all files (for example Rollup does this - https://github.com/rollup/rollup/blob/c0c206e2a8a892fdb0e4f9b25b0a817b733a072f/rollup.config.js#L118 )

thysultan: >We can, however, use .cjs to avoid this extra directory.

If we can avoid this, let us export a .cjs file, i think rollup can produce a dedicated cjs file without the wrappers involved in the umd file.

Andarist: Ok, added a separate config item for `.cjs` file

script/build.js
```diff
@@ -1,8 +1,8 @@
 import {join} from 'path'
-import {terser} from "rollup-plugin-terser"
+import {terser} from 'rollup-plugin-terser'
 import size from 'rollup-plugin-size'
 
-const options = {mangle: true, compress: false}
+const options = {mangle: true, compress: false, toplevel: true}
```
thysultan, line 5: What does the toplevel option do?

thysultan: >--toplevel                  Compress and/or mangle variables in top level scope.

NVM

thysultan:
LGTM

package.json
thysultan: We still want to point main to the umd file.

Andarist: Just because conceptually I see `pkg.main` as `pkg.exports.require`, so I've reused the same file - it also avoids UMD wrapper for node@<13 consumers, not a big deal - can change it back if you want.

PR #208, opened 2020-04-23
Automatic d.ts generation from jsdoc

SalvatorePreviti:
Adds automatic d.ts generation to make the package usable in typescript.

- Adds types/ to .gitignore
- Adds tsconfig.json to generate d.ts from js files
- Adds "types": "types/" in package.json
- Adds typescript devDependency
- Adds "tsc &&" before rollup

Tested by linking the package in a typescript project.

Thank you :)


thysultan:
Can you commit the `types/` files so we can see how good it is?

SalvatorePreviti:
Yes, give me some minutes

SalvatorePreviti:
Consider that optional arguments at the moment are not optional because jsdoc way to mark a parameter optional is @param {type} [argument], with the squared bracket. But types are still usable enough

thysultan:
What's the difference between this and TypeScript using the jsdocs directly?

SalvatorePreviti:
Not all projects can have "allowJs" enabled and/or don't compile if d.ts are not provided with a "types missing". This force the people to add a fake .d.ts with module "stylis" that just exports any

SalvatorePreviti:
Whilst this definition does not cover all cases with strong typing, is still a base start point to add types. Also, refactoring the jsdoc a bit would in the future provide better types than leaving the package without this

thysultan:
Is there a way to encode the `tsconfig.json` config in `package.json` instead of introducing another config file?

Andarist:
> Is there a way to encode the tsconfig.json config in package.json

I don't believe there is.

> or some combo of package.json. and script.build.js instead of introducing another config file?

This, on the other hand, should be possible - by using some Rollup's TS plugin (there are like 3, lol) you could just feed options to it directly rather than relying on tsconfig.json. It comes with a caveat though - code editor's (such as VSCode) would not be aware of such a config, so we'd miss on all benefits from having the project TS-aware while working on the repo. It would only be beneficial to Stylis consumers.

thysultan:
>It comes with a caveat though - code editor's (such as VSCode) would not be aware of such a config

I thought they use the `pkg.typings` directory for typings, what would tsconfig.json serve beyond serving the typings generated in `/types`?

Andarist:
Screenshot 2020-04-27 at 23 24 40
Even though we get some IntelliSense (I can see parameter types etc), we don't get type-checking so I might misuse the existing (and typed~) APIs to only realize I've done a mistake later.

> what would tsconfig.json serve beyond serving the typings generated in /types?

tsconfig.json is not relevant for this at all, it only allows generating those types (and other TS features while working on the project), but is not needed for consuming generated types at all. Like you have mentioned - for consuming types you only need `pkg.typings` (or `pkg.types` or just a sibling `.d.ts` file to your `.js`file)

SalvatorePreviti:
No is not possible to put tsconfig in package.json, however, you can put it in a subfolder and use `tsc -p folder/tsconfig.json` and can be renamed in the way you prefer.
Tsconfig.json does not need to be included in the final package so is going to be used only to generate the .d.ts before releasing

SalvatorePreviti:
Would you like me to move tsconfig.json to script/tsconfig.build.json folder?

heyheyhello:
Has this not been merged because you'd rather see the types in something like DefinitelyTyped or @types/stylis?

SalvatorePreviti:
It would be easy to do it also in the same repository, of course - generate the .d.ts file in a "typings" folder that contains another package.json (folder not published with the .js package) and in the publish script is possible to publish both packages

thysultan:
Per its inclusion in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46171

PR #215, opened 2020-05-01
fix: compatibility with Rollup

kripod:
[`@rollup/plugin-node-resolve`](https://github.com/rollup/plugins/tree/master/packages/node-resolve) works out of the box only when the `main` field is set to the CommonJS bundle.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,11 +9,13 @@
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "sideEffects": false,
   "type": "module",
-  "main": "dist/stylis.umd.js",
   "module": "dist/stylis.esm.js",
+  "main": "dist/stylis.cjs",
+  "umd:main": "dist/stylis.umd.js",
   "exports": {
     "import": "./index.js",
-    "require": "./dist/stylis.cjs"
+    "require": "./dist/stylis.cjs",
+    "umd": "dist/stylis.umd.js"
   },
   "files": [
     "index.js",


kripod:
It turns out that using [`@rollup/plugin-commonjs`](https://github.com/rollup/plugins/tree/master/packages/commonjs) seems to resolve the issue. Sorry for the nuisance!

Andarist:
That's an interesting case though - could you check if maybe there is already an issue about conditional exports not being supported by `@rollup/plugin-node-resolve` and if not, could you create one?

kripod:
Thanks for the heads-up! I've opened https://github.com/rollup/plugins/issues/362 as I couldn't find any previous issues or pull requests regarding the case.

dburles:
Hey, any idea when this change might be published? Currently stuck with failing builds as Node 12 dumps an error due to the way the exports were defined previously.

Andarist:
@dburles AFAIK there is nothing to be published here. You should just use `@rollup/plugin-commonjs` or help fixing https://github.com/rollup/plugins/issues/362 . Stylis itself is correctly published - if you don't believe that is the case then please share a runnable repro case which I could examine.

dburles:
Hey @Andarist, I am using stylis as a dependency of a package for vendor prefixing, here's a link to the failure: https://github.com/dburles/mystical-alpha/runs/652169304

Andarist:
@dburles I had an idea what this could be but unfortunately, that was not it ( https://github.com/dburles/mystical-alpha/pull/2 ). The problem is somewhat weird I can't reproduce it locally and node@12 shouldn't try to interpret `package.json#type` etc without a flag and I can't see the flag being passed anywhere to it, so 🤷 Not sure how to debug this further right now.

dburles:
Thanks for taking a look into it, how strange. I’ll try to have a look into it.

dburles:
Basically Node isn't happy with `type: "module"` being declared and instead finding cjs. Is it being declared just for bundlers?

Andarist:
> Is it being declared just for bundlers?

No, it is declared for node - it tells how to interpret `.js` files in a package. Node should try to read this only if:
- you are using an experimental flag
- you are using node which has unflagged support for ESM modules (which v12 does not). This has been only unflagged in node@13.2 - see [changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#2019-11-21-version-1320-current-mylesborins).

PR #228, opened 2020-06-25
Change package.json#main to `.cjs` file

Andarist:
It seems that node@12.16.3 ships with unflagged module support BUT without support for conditional exports, so it just tries to use `"main"` but at the same type it sees `"type": "module"` and fails on `require`ing Stylis because it thinks it's going to load a module and it can't do that when using `require`.

This node version is used by Vercel and AWS, so it's pretty important for us to support them. The issue can be reproduced with this version of node using this repository: https://github.com/harrisrobin/vercel-deploy-issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "pretest": "npm run lint && npm run build",
     "test": "nyc npm run spec",
     "spec": "mocha --harmony --require esm script/setup.js --recursive test",
+    "prebuild": "rimraf dist",
     "build": "rollup --config script/build.js --configSrc ./",
     "start": "rollup --config script/build.js --configSrc ./ --watch",
     "prepare": "npm run build",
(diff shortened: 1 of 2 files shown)


thysultan:
We could also change the `umd` file to just `stylis.cjs` and remove the `cjs` file since it can serve both.

Andarist:
@thysultan sure, that works too - I've pushed out the change. Please remember about clearing `dist` before releasing a new version so there is no leftover UMD file from previous builds.

thysultan:
>Please remember about clearing dist before releasing a new version

Can you add that to the build process? i.e before generating the dist files. Does rollup provide an opt for this?

Andarist:
Rollup itself doesn't have an option for this. I usually just use `rimraf` (which works on all platforms, including Windows) for this and thus I've added it here.

thysultan:
When does `prebuild` run? or is that a rollup thing?

Andarist:
Its npm scripts, you can add pre and post scripts for any other script. It works with all packaga managers

thysultan:
We can change the "start" script to "npm run build -- --watch" so it can benefit from this as well.

thysultan:
Other than that LGTM.

Andarist:
Done.

PR #231, opened 2020-07-14
Better JSDoc types

heyheyhello:
I work almost exclusively in Typescript and not having type support is a show stopper for what is otherwise a lovely project. I looked at #208 and while I think it has good intentions the types are vague. That's not the PR author's fault - `tsc` just extracts the types from JSDoc. The JSDoc typings could be better.

I went through and _tried_ to fix it up. I removed all the instances of `object` and `function` with their real types. There are a few errors and maybe you can correct me on some assumptions I had to make about what is a possible data type but I think these are good types - specifically the use of `Middleware` as a type to explain that a function _is_ middleware (or returns middleware in the case of `rulesheet()`).

I arbitrarily picked Middleware.js to write the JSDoc `@typedef` but they could be anywhere - maybe index.js would be a better place?

I'll leave comments in the PR code to explain what I'm trying to do

### DIFF
--- a/src/Middleware.js
+++ b/src/Middleware.js
@@ -5,31 +5,118 @@ import {serialize} from './Serializer.js'
 import {prefix} from './Prefixer.js'
 
 /**
- * @typedef {{
- * 	 parent?: Element,
- * 	 children?: Element[] | string,
- * 	 root: Element,
- * 	 type: string,
- * 	 props: string[] | string,
- * 	 value: string,
- * 	 length: number,
- * 	 return: string,
- * 	 line?: number,
- * 	 column?: number,
- * }} Element
+ * @typedef {Object} NodeBase
+ * @property {Node?} parent
+ * @property {Node?} root
+ * @property {string} value
+ * @property {number} length
+ * @property {string} return
+ * @property {number} line
+ * @property {number} column
+
+ * @typedef {NodeBase & {
+	type: 'comm'
+	props: string
+	children: string
+ }} CommentNode
+
+ * @typedef {NodeBase & {
+	type: 'decl'
+	props: string
+	children: string
+ }} DeclarationNode
+
+ * @typedef {NodeBase & {
+	type: 'rule'
+	props: string[]
+	children: Node[]
+ }} RulesetNode
+
+ * @typedef {NodeBase & {
+	type: '@keyframes'
+	props: string[]
+	children: Node[]
+ }} KeyframesNode
+
+* @typedef {NodeBase & {
+	type: '@media'
+	props: string[]
+	children: Node[]
+ }} MediaNode
+
+* @typedef {NodeBase & {
+	type: '@supports'
+	props: string[]
+	children: Node[]
+ }} SupportsNode
+
+ * @typedef {NodeBase & {
+	type: '@import'
+	props: []
+	children: []
+ }} ImportNode
+
+ * @typedef {NodeBase & {
+	type: '@charset'
+	props: []
+	children: []
+ }} CharsetNode
+
+* @typedef {NodeBase & {
+	type: '@counter-style'
+	props: string[]
+	children: DeclarationNode[]
+ }} CounterStyleNode
+
+ * @typedef {NodeBase & {
+	type: '@font-face'
+	props: []
+	children: DeclarationNode[]
+ }} FontFaceNode
+
+ * @typedef {NodeBase & {
+	type: '@page'
+	props: []
+	children: DeclarationNode[]
+ }} PageNode
+
+ * @typedef {NodeBase & {
+	type: '@document'
+	props: string[]
+	children: Node[]
+ }} DocumentNode
+
+ * @typedef {NodeBase & {
+	type: '@viewport'
+	props: []
+	children: DeclarationNode[]
+ }} ViewportNode
+
+ * @typedef {NodeBase & {
+	type: '@-webkit-keyframes'
+	props: string[]
+	children: Node[]
+ }} WebkitKeyframesNode
+
+ * @typedef {NodeBase & {
+	type: '@-moz-document'
+	props: string[]
+	children: Node[]
+ }} MozDocumentNode
+
+ * @typedef {NodeBase & {
+	type: '@-ms-viewport'
+	props: []
+	children: DeclarationNode[]
+ }} MsViewportNode
+
+ * @typedef {CommentNode | DeclarationNode | RulesetNode | KeyframesNode | MediaNode | SupportsNode | ImportNode | CharsetNode | CounterStyleNode | FontFaceNode | ViewportNode | PageNode | DocumentNode | WebkitKeyframesNode | MozDocumentNode | MsViewportNode} Node
 
  * @typedef {(
- * 	 value: string,
+ * 	 element: Node,
  * 	 index: number,
- * 	 array: string[],
- * 	) => string
- * } ArrayMapCallback
-
- * @typedef {(
- * 	 element: Element,
- * 	 index?: number,
- * 	 children?: (Element | string)[],
- * 	 callback?: Middleware,
+ * 	 children: Node[],
+ * 	 callback: Middleware,
  * 	) => string | void
  * } Middleware
  */
@@ -52,13 +139,15 @@ export function middleware (collection) {
 }
 
 /**
- * @param {(ret: string) => void} callback
+ * @param {(rule: string) => void} callback
  * @return {Middleware}
  */
 export function rulesheet (callback) {
 	return function (element) {
 		if (!element.root)
+			// @ts-ignore
 			if (element = element.return)
+				// @ts-ignore
 				callback(element)
 	}
 }
@@ -118,6 +207,7 @@ export function namespace (element) {
 							return index === 1 ? '' : value
 						default:
 							switch (index) {
+								// @ts-ignore
 								case 0: element = value
 									return sizeof(children) > 1 ? '' : value
 								case index = sizeof(children) - 1: case 2:
--- a/src/Middleware.js
+++ b/src/Middleware.js
@@ -5,8 +5,36 @@ import {serialize} from './Serializer.js'
 import {prefix} from './Prefixer.js'
 
 /**
- * @param {function[]} collection
- * @return {function}
+ * @typedef {{
+ * 	children?: Element[],
+ * 	length: number,
+ * 	props: string[],
+ * 	return: string,
+ * 	root: Element,
+ * 	parent?: Element,
+ * 	type: string,
+ * 	value: string,
+ * }} Element
+
+ * @typedef {(
+ *	 value: string,
+ *	 index: number,
+ *	 array: string[],
+ *	) => string
+ * } Callback
+
+ * @typedef {(
+ * 	 element: Element,
+ * 	 index?: number,
+ * 	 children?: Element[],
+ * 	 callback?: Middleware,
+ * 	) => string | void
+ * } Middleware
+ */
+
+/**
+ * @param {Middleware[]} collection
+ * @return {Middleware}
  */
 export function middleware (collection) {
 	var length = sizeof(collection)
@@ -22,22 +50,19 @@ export function middleware (collection) {
 }
 
 /**
- * @param {function} callback
- * @return {function}
+ * @param {(ret: string) => void} callback
+ * @return {Middleware}
  */
 export function rulesheet (callback) {
 	return function (element) {
 		if (!element.root)
 			if (element = element.return)
-				callback(element)
+				callback(/** @type {string} */ (element))
 	}
 }
 
 /**
- * @param {object} element
- * @param {number} index
- * @param {object[]} children
- * @param {function} callback
+ * @type {Middleware}
  */
 export function prefixer (element, index, children, callback) {
 	if (!element.return)
@@ -68,9 +93,7 @@ export function prefixer (element, index, children, callback) {
 }
 
 /**
- * @param {object} element
- * @param {number} index
- * @param {object[]} children
+ * @type {Middleware}
  */
 export function namespace (element) {
 	switch (element.type) {
--- a/src/Middleware.js
+++ b/src/Middleware.js
@@ -6,27 +6,29 @@ import {prefix} from './Prefixer.js'
 
 /**
  * @typedef {{
- * 	children?: Element[],
- * 	length: number,
- * 	props: string[],
- * 	return: string,
- * 	root: Element,
- * 	parent?: Element,
- * 	type: string,
- * 	value: string,
+ * 	 parent?: Element,
+ * 	 children?: Element[] | string,
+ * 	 root: Element,
+ * 	 type: string,
+ * 	 props: string[] | string,
+ * 	 value: string,
+ * 	 length: number,
+ * 	 return: string,
+ * 	 line?: number,
+ * 	 column?: number,
  * }} Element
 
  * @typedef {(
- *	 value: string,
- *	 index: number,
- *	 array: string[],
- *	) => string
- * } Callback
+ * 	 value: string,
+ * 	 index: number,
+ * 	 array: string[],
+ * 	) => string
+ * } ArrayMapCallback
 
  * @typedef {(
  * 	 element: Element,
  * 	 index?: number,
- * 	 children?: Element[],
+ * 	 children?: (Element | string)[],
  * 	 callback?: Middleware,
  * 	) => string | void
  * } Middleware
--- a/src/Middleware.js
+++ b/src/Middleware.js
@@ -57,7 +57,7 @@ export function rulesheet (callback) {
 	return function (element) {
 		if (!element.root)
 			if (element = element.return)
-				callback(/** @type {string} */ (element))
+				callback(element)
 	}
 }
 
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -2,27 +2,30 @@ import {COMMENT, RULESET, DECLARATION} from './Enum.js'
 import {abs, trim, from, sizeof, strlen, substr, append, replace} from './Utility.js'
 import {node, char, next, peek, caret, alloc, dealloc, delimit, whitespace, identifier, commenter} from './Tokenizer.js'
 
-/** @typedef {import('./Middleware.js').Element} Element */
+/** @typedef {import('./Middleware.js').Node} Node */
+/** @typedef {import('./Middleware.js').CommentNode} CommentNode */
+/** @typedef {import('./Middleware.js').DeclarationNode} DeclarationNode */
 
 /**
  * @param {string} value
- * @return {Element[]}
+ * @return {Node[]}
  */
 export function compile (value) {
+	// @ts-ignore
 	return dealloc(parse('', null, null, null, [''], value = alloc(value), 0, [0], value))
 }
 
 /**
  * @param {string} value
- * @param {Element} root
- * @param {Element?} parent
- * @param {Element[]} rule
+ * @param {Node?} root
+ * @param {Node?} parent
+ * @param {Node?} rule
  * @param {string[]} rules
- * @param {Element[]} rulesets
+ * @param {Node[]} rulesets
  * @param {number} pseudo
  * @param {number[]} points
- * @param {string[]} declarations
- * @return {Element[]}
+ * @param {Node[]} declarations
+ * @return {Node[]}
  */
 export function parse (value, root, parent, rule, rules, rulesets, pseudo, points, declarations) {
 	var index = 0
@@ -87,6 +90,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 								switch (atrule) {
 									// d m s
 									case 100: case 109: case 115:
+										// @ts-ignore
 										parse(value, reference, reference, rule && append(ruleset(value, reference, reference, 0, 0, rules, points, type, rules, props = [], length), children), rules, children, length, points, rule ? props : children)
 										break
 									default:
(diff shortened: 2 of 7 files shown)


src/Parser.js
```diff
@@ -2,25 +2,29 @@ import {COMMENT, RULESET, DECLARATION} from './Enum.js'
...
+
 /**
  * @param {string} value
- * @return {object[]}
+ * @return {Rulesets}
  */
 export function compile (value) {
-	return dealloc(parse('', null, null, null, [''], value = alloc(value), 0, [0], value))
+	alloc(value)
+	return dealloc(parse('', null, null, null, [''], [], 0, [0], []))
```
heyheyhello, line 14: Is this what you're trying to do?

src/Tokenizer.js
```diff
@@ -7,22 +7,24 @@ export var position = 0
...
+ * @param {Element?} parent
  * @param {string} type
  * @param {string[]} props
- * @param {object[]} children
+ * @param {Element[]} children
  * @param {number} length
  */
 export function node (value, root, parent, type, props, children, length) {
-	return {value: value, root: root, parent: parent, type: type, props: props, children: children, line: line, column: column, length: length, return: ''}
+	return {value, root, parent, type, props, children, line, column, length, return: ''}
```
heyheyhello, line 22: Do we need ES5 support?

src/Tokenizer.js
```diff
@@ -101,15 +103,16 @@ export function token (type) {
...
 export function alloc (value) {
 	return line = column = 1, length = strlen(characters = value), position = 0, []
 }
 
 /**
- * @param {any} value
- * @return {any}
+ * @typedef {any} T
+ * @param {T} value
+ * @return {T}
```
heyheyhello, line 115: I'm trying to show that what you give it is what you get back, as `T`

src/Utility.js
```diff
@@ -30,10 +32,11 @@ export function trim (value) {
 /**
  * @param {string} value
  * @param {RegExp} pattern
- * @return {string?}
+ * @return {string}
```
heyheyhello, line 35: Never null?

src/Parser.js
```diff
@@ -2,25 +2,29 @@ import {COMMENT, RULESET, DECLARATION} from './Enum.js'
...
 }
 
 /**
  * @param {string} value
- * @param {object} root
- * @param {object?} parent
- * @param {string[]} rule
+ * @param {Element} root
+ * @param {Element?} parent
+ * @param {Element | string[]} rule
```
heyheyhello, line 21: `rule` looks like an Element sometimes

heyheyhello:
If these look good I'll compile a single index.d.ts of all types and upload it to npm as `@types/stylis` unless you want it in this repo

src/Tokenizer.js
thysultan: Yes.

src/Tokenizer.js
```diff
@@ -101,15 +103,16 @@ export function token (type) {
 
 /**
  * @param {string} value
- * @return {any[]}
+ * @return {[]}
```
Andarist, line 106: won't this be an equivalent to `never[]` or something? from what I remember this allocates `Element[]`, I think this wouldn't compile correctly when declared like this when using TS, not sure what about JSDoc though

src/Tokenizer.js
Andarist: this could use `@template T`

src/Parser.js
```diff
@@ -2,25 +2,28 @@ import {COMMENT, RULESET, DECLARATION} from './Enum.js'
 import {abs, trim, from, sizeof, strlen, substr, append, replace} from './Utility.js'
 import {node, char, next, peek, caret, alloc, dealloc, delimit, whitespace, identifier, commenter} from './Tokenizer.js'
 
+/** @typedef {import('./Middleware.js').Element} Element */
+/** @typedef {string[]} Rulesets */
```
Andarist, line 6: IMHO it's better to declare `Ruleset` and use `Ruleset[]` later. But anyway - I'm not sure if this is declared correctly, this type is used as a return type of the `compile` which returns `Element[]` - it's an array of objects, not an array of strings

src/Utility.js
Andarist: notice how you have changed what this function does, previously `value` was reassigned as the code was written to be as terse as possible, so it was actually possible that we could have returned `null` from it as `.match` call might return it. Right now you are returning the input value and this is not an equivalent

src/Utility.js
```diff
@@ -91,17 +94,18 @@ export function sizeof (value) {
 }
 
 /**
- * @param {any} value
+ * @typedef {any} T
```
Andarist, line 97: Similarly, you could use `@template` here

src/Tokenizer.js
heyheyhello: Hmm. It returns _exactly_ `[]`. Not sure what it's going to be used for afterwards but it's `[]` at this point

src/Parser.js
heyheyhello: OK good to know - it's not clear that `compile` is returning `Element[]` because `compile` is `dealloc` (which returns what it is given) after a `parse` (which ends in `return rulesets`). Are you suggesting that `rulesets` in `parse` is `Element[]`?

I like the note about using `Ruleset` and `Ruleset[]`. I'll do that thanks

heyheyhello:
@Andarist Thanks for all the feedback! That's very helpful. I went and pushed your suggested changes about `@template` (wish I knew that existed sooner), `Ruleset[]`, and not realizing the behaviour change in `match()`.

As for changing the codebase to Typescript, I don't think it's worth the fuss. @thysultan doesn't seem to already use TS in any way, and didn't seem familiar with it in #208. It would add unnecessary complexity to an otherwise simple and small project that can be adequately types by JSDoc. Furthermore, the author likes to use very terse code in a handful of places that Typescript just wouldn't allow - you'd have to either rewrite it (like I was trying to do with `match()` earlier) or use a bunch of type assertions like `value = element as unknown as Element`. I don't think it's worth the headache.

heyheyhello:
Converting a project to TS is only ever for the author anyway. Users of a library will always need an index.js+index.d.ts. I think it's more important to get that d.ts out the door so people can actually use this lib with their TS code. I'll upload to @types/stylis as soon as we can confirm the emitted declaration from JSDoc is sufficient

heyheyhello:
Will be merging this: https://github.com/DefinitelyTyped/DefinitelyTyped/compare/master...heyheyhello:master

src/Middleware.js
```diff
@@ -6,27 +6,29 @@ import {prefix} from './Prefixer.js'
...
+ * 	 index: number,
+ * 	 array: string[],
+ * 	) => string
+ * } ArrayMapCallback
 
  * @typedef {(
  * 	 element: Element,
  * 	 index?: number,
- * 	 children?: Element[],
+ * 	 children?: (Element | string)[],
```
heyheyhello, line 31: See readme.md where apparently this can be a string

src/Middleware.js
```diff
@@ -6,27 +6,29 @@ import {prefix} from './Prefixer.js'
...
- * 	return: string,
- * 	root: Element,
- * 	parent?: Element,
- * 	type: string,
- * 	value: string,
+ * 	 parent?: Element,
+ * 	 children?: Element[] | string,
+ * 	 root: Element,
+ * 	 type: string,
+ * 	 props: string[] | string,
```
heyheyhello, line 13: Similarly, this is apparently a string in some cases. Though all references in the codebase treat it as an array (using methods like `.join()`)

heyheyhello:
PR is now open at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46171

@Andarist would you like to be reviewer and/or co-owner as per [bot comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46171#issuecomment-660522205)

> bell @heyheyhello — you're the only owner, but it would still be good if you find someone to [review this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46171/files) in the next few days, otherwise a maintainer will look at it. (And if you do find someone, maybe even recruit them to be a second owner to make future changes easier...)

src/Tokenizer.js
heyheyhello: I think that's a limitation of the way the code is written - If this package *had any documentation* it'd be good for it to go in there. I agree that it's confusing, but I think a lot of this codebase is. A lot of functions have side effects on shared global data, or are aliases to other functions, or change data types randomly in order to have "terse" code. It feels like there's too much to fix in terms of type safety.

I'm making another comment in this thread in a sec, basically saying I'd be happy for you to take over since I imagine this PR will stall.

src/Tokenizer.js
heyheyhello: Ok fair. I'll use `unknown[]`. Good call

src/Utility.js
```diff
@@ -91,17 +93,18 @@ export function sizeof (value) {
...
+ * @return {T}
  */
 export function append (value, array) {
 	return array.push(value), value
 }
 
 /**
  * @param {string[]} array
- * @param {function} callback
+ * @param {ArrayMapCallback} callback
```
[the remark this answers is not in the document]
heyheyhello: It's mostly to differentiate from Middleware which looks _super similar_ and middleware is sometimes passed around as the variable `callback` 🙄

I agree that the type is not as generic as it could be, but it's actually only used **twice** in the whole codebase and my preference would be to *not* have `combine` at all - use `[].map().join()`. The author likes to have a lot of aliases though, and I don't think it's worth trying to convince them to change that

src/Middleware.js
thysultan: Can we put the typedef's(single line) where they are created in the code i.e for the Element typedef we can call it node is it is returned from the `node` function and put it next to the `node` functions return JSDOC. For example in https://github.com/thysultan/stylis.js/blob/eff8dc8a0ca05c3a80054ca57915e8a5dba0ab9e/src/Tokenizer.js#L21:

```
/**
 * @typedef {{parent?: node, children?: node[] | string, root: node, type: string, props: string[] | string, value: string, length: number, return: string, line?: number, column?: number}} node
 * @param {string} value
 * @param {nod} root
 * @param {node?} parent
 * @param {string} type
 * @param {string[]} props
 * @param {node[]} children
 * @param {number} length
 * @return node
 */
```

Same thing for middleware in https://github.com/thysultan/stylis.js/blob/eff8dc8a0ca05c3a80054ca57915e8a5dba0ab9e/src/Middleware.js#L41

```
/**
 * @typedef {(element: node, index?: number, children?: (node | string)[], callback?: middleware) => string | void} middleware
 * @param {middleware[]} collection
 * @return {middleware}
 */
```

thysultan:
> add `jsconfig.json` to the project (or `tsconfig.json)`.

I thought this was only the case for the consumer?

heyheyhello:
> I would like to wait a couple of days for @thysultan opinion about this before moving this work to DefinitelyTyped. From my POV it would be better to have those types defined here (even if using JSDoc) to avoid desynchronization of the codebase and types.

I agree that it's normally ideal for types to exist in the source codebase, but I think its unrealistic in this case. #208 has been open for months. There's not much show of interest from @thysultan about moving forward and I don't blame them - having d.ts in the codebase would be a huge overhaul. The types I have in DT aren't reproducible from `tsc` as done in 208 - there's actually a small handful of manual edits needed.   

> I believe we should try to typecheck the codebase to assess that what has been typed is in fact correct - and to do that I think we should add jsconfig.json to the project (or tsconfig.json). At least it's the only thing that made VScode to start typechecking this for me.

image

This will do type checking. But you'll notice that the codebase is not written with type safety in mind - even the nature of having helper functions like `combine` and `strlen` make types awkward. There are a lot of errors.

> I've checked out your branch and fixed a bunch of minor issues - there are some other type problems though and I will have to look into them a little bit later.

I just don't think the author wants type problems to be fixed or the project to move more towards TypeScript (or be ported to TS). @thysultan can you comment on your interest or how important it is to you to re-write in that direction?

#208 is months old and I know there are people who wanted types available for this as soon as 4.0 was out.

I'll push to DefinitelyTyped, and if this ever gets merged or type support becomes first class in this codebase I'll be happy to remove the types from DT as documented in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#removing-a-package

thysultan:
>@thysultan can you comment on your interest or how important it is to you to re-write in that direction?

JSDocs yes, not so much a TypeScript re-write. See my comments on my suggestions on this PR.

src/Tokenizer.js
Andarist: In a similar fashion `unknown[]` is not a good fit here because it would require either casting or type refining down the road - we want the result of `alloc` to be consumable by other code. That's why generic (`T[]`) would be preferred here.

Look what happens with `unknown`:
https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABAQwDarhAFASgFyLgDWYcA7mANoC6iA3gFCLOIBOAplCK0jQwL4MGUAJ4AHdogBKIVJIC89URLwByVrPaqANIgBuaEOwIBnKKxhgA5oIAm7CKmQdEEBGbYEZcoQdYp0THZbREU0DGwcBnCg2wA6MRATAAssViiGNzAPVgAmL01QgIjgygAGaiEgA

src/Middleware.js
Andarist: We want `Node` (or `Element`, whatever we are going to have it called) to be reusable across different declarations, not sure how defining a `@typedef` within the same block as a function signature would work. Even if it would work it kinda blends together for me and I wouldn't call this more readable in the end of a day. Defining different types of nodes (`CommentNode`, `DeclarationNode`, etc) is also worth exploring - so I think keeping them separate (but probably in the same file) would improve readability.

Andarist:
> > add jsconfig.json to the project (or tsconfig.json).

> I thought this was only the case for the consumer?

The ultimate goal of this work is to both document the codebase better and to generate `.d.ts` files automatically. We have to call TS compiler for both and I would rather say that it's better to have it called with explicit settings rather than some defaults (if it even could be called without explicit config). I have only managed to get VScode to typecheck the codebase based on those JSDocs after adding this `jsconfig.json` file:
Screenshot 2020-07-19 at 21 54 12

> This will do type checking. But you'll notice that the codebase is not written with type safety in mind - even the nature of having helper functions like combine and strlen make types awkward. There are a lot of errors.

Yeah, I realize the codebase is using some unsafe, from the perspective of the types, patterns but sprinkling few `@ts-ignore` here and there should get us pretty far. Or at least that's my hope.

> I'll push to DefinitelyTyped, and if this ever gets merged or type support becomes first class in this codebase I'll be happy to remove the types from DT as documented in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#removing-a-package

I think it would be better to close that DefinitelyTyped PR for like a week or two. Let us figure this out here in that timeframe and if we won't be able to get the job done you will be able to reopen that PR there.

src/Middleware.js
thysultan: >Defining different types of nodes (CommentNode, DeclarationNode, etc)

That might be too much to represent in JSDocs given the constraints of jsdocs and my reluctants to spear head a full-on TypeScript re-write. If it works i think an inline @typedef has the benefit of highlighting where to the type is sourced/created from avoiding indirection; in contrast: in its free-floating form it raises the question of whether it's from global source.

heyheyhello:
😍

Nice! It's a lot of code but looks good to me. I imagine it's a bit overwhelming to anyone not familiar with types but certainly a better way to go.

src/Middleware.js
```diff
@@ -52,13 +139,15 @@ export function middleware (collection) {
...
 
 /**
- * @param {(ret: string) => void} callback
+ * @param {(rule: string) => void} callback
  * @return {Middleware}
  */
 export function rulesheet (callback) {
 	return function (element) {
 		if (!element.root)
+			// @ts-ignore
```
thysultan, line 148: Not a fan of the ts-ignore comments, considering this commit essentially create a full type-definition we could just shop a `typings.d.ts` definition file with this in it instead of JSDoc comments?

src/Utility.js
```diff
@@ -1,14 +1,10 @@
-/** @typedef {import('./Middleware.js').ArrayMapCallback} ArrayMapCallback */
-
 /**
- * @param {number}
```
Andarist, line 2: old definitions were not picked up by the IDE at all - I think those can only be used for actual function declarations and not for defining types when assigning to a variable like this

src/Tokenizer.js
```diff
@@ -7,28 +7,33 @@ export var position = 0
...
  */
 export function node (value, root, parent, type, props, children, length) {
+	// @ts-ignore
 	return {value: value, root: root, parent: parent, type: type, props: props, children: children, line: line, column: column, length: length, return: ''}
 }
 
 /**
  * @param {string} value
- * @param {Element} root
+ * @param {Node} source
```
Andarist, line 30: i've renamed this param as `root` was IMHO a confusing name for this - given that we have a `.root` property on Nodes

src/Parser.js
```diff
@@ -129,17 +133,17 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
...
  * @param {string} value
- * @param {Element} root
- * @param {Element?} parent
+ * @param {Node?} root
+ * @param {Node?} parent
  * @param {number} index
  * @param {number} offset
  * @param {string[]} rules
  * @param {number[]} points
  * @param {string} type
```
Andarist, line 142: potentially we could explore limiting this to a particular union of strings and define the whole function in a similar fashion as [`node`](https://github.com/thysultan/stylis.js/blob/ceb1b0416fab3b4c451334b35c64a62529ab91dd/src/Tokenizer.js#L12-L23)

heyheyhello:
@Andarist I think your changes would be great as a PR to update the typings on DefinitelyTyped. The typings that are available+published now as @types/stylis work well for developers using stylis, but your changes would benefit people trying to write middleware and hack on stylis itself. Specifically adding node subtypes and renaming parameters (i.e root to source) would benefit developers who are working with stylis internals - which may be groups like emotion and styled-components (who are currently stuck on V3).

If you're interested in opening a PR there I'll approve them and the tests can be edited to cover more type-specific usecases like writing middleware.

Andarist:
Sorry for being MIA lately - had a lot on my shoulders and couldn't find time to get back to this PR.

@heyheyhello I think we have every intention of shipping those types (after some redacting) as part of the Stylis itself. We just need to figure out how we want to do this - which shouldn't take us **that** long. I hoped you wouldn't rush with this DefinitelyTyped PR too much as now we'll be in this weird situation where types will be published (soon-ish) both within `stylis` and `@types/stylis` packages which will only confuse some users.

src/Middleware.js
Andarist: > Not a fan of the ts-ignore comments

Plan to remove them, they were only somewhat helpful to me when I've been typing this.

> considering this commit essentially create a full type-definition we could just shop a typings.d.ts definition file with this in it instead of JSDoc comments?

Yeah, we can totally do that - my idea was just that the types were sort-of already here, so I've tweaked those instead of providing a separate set of types. Would you like me to bring back old&simple JSDocs and gather all of those types within `.d.ts` file?

heyheyhello:
@Andarist There's no issue with that. It's something that DefinitelyTyped plans for and has deprecation warnings for. It's easy to open a PR in DT to have the types taken down :) If you shipped types tomorrow no one installing @types/stylis would see an issue; the typings would still resolve. 

I didn't feel rushed with DT, but I've seen many PRs stay open for months or _years_ so why not have the ball rolling now? I didn't see any reason to stop progress

Andarist:
What will happen with people who have already installed `@types/stylis` and who will upgrade to a new Stylis with types included in it? It's a genuine question - I'm not sure which one wins in such a situation or if they cause a conflict or not.

heyheyhello:
I'm pretty sure it's always setup to prefer local types. I just checked too:
image

I copy/pasted the node_modules/@types/stylis into node_modules/stylis/ and added the `"types": "index.d.ts"` to package.json, and then I comment out the `serialize` method just to be sure it updates. Works as expected.

heyheyhello:
> I didn't feel rushed with DT, but I've seen many PRs stay open for months or _years_ so why not have the ball rolling now?

Also a huge push for me is that my memory's not great - it's been 20 days since this PR was opened and honestly I don't remember writing the types. There's a small window before I move onto other things...

heyheyhello:
Thought I'd mention now that I've been using @types/stylis for a bit... The huge amount of commonly named exports that stylis has is problematic: things like `next`, `length`, and `substr`. My editor autocomplete/autoimport messes up when I type `nex<Enter>` hoping to resolve "nextParent" (my variable name) but it auto-imports `next` from stylis :unamused: 

Not sure the best way forward. If not remove exports then how about renaming or prefixing?

Andarist:
@heyheyhello we can't just rename them in v4 as that would be a breaking change, so even if we would have provided aliases for those the problem for you would still persist. It really feels like your IDE's algorithm could be improved to prefer local variables? Can't it be configured somehow? I also don't see this much of a problem - one library's API choices can't be dictated by IDEs behaviors.

thysultan:
@heyheyhello Why would it try to autocomplete for exports you haven't imported? unless you're importing everything?

```js
import * as stylis from 'stylis'

stylis.next(...)
```

heyheyhello:
> @heyheyhello Why would it try to autocomplete for exports you haven't imported? unless you're importing everything?
> 
> ```js
> import * as stylis from 'stylis'
> 
> stylis.next(...)
> ```


That's a great idea thanks. Star import is the way to go. Also my editor does auto imports. I think vscode offers it by default? So importing only one thing from stylis tells the editor to be ready to offer all symbols and just patch the import statement as needed.

Sorry I thought it was a common feature. Good fix!

heyheyhello:
Hey I'm tired of seeing this on my homepage so I'll close this. I've already done all the changes to @types/stylis and don't want this to hang forever like #208. Life's too short. Thanks for all the help with it @Andarist! Looking forward to any future PRs that propose better types.

MaxMilton:
@heyheyhello great work on `@types/stylis`, thanks for your efforts! I had been using my own until I noticed it in your last comment.

Shame about the poorly typed JSDoc annotations... but now with `@types/stylis` it's not an issue for the lib consumers :)

PR #232, opened 2020-07-14
Update packages

heyheyhello:
Builds OK :)


package.json
```diff
@@ -25,7 +25,7 @@
     "pretest": "npm run lint && npm run build",
     "test": "nyc npm run spec",
     "spec": "mocha --harmony --require esm script/setup.js --recursive test",
-    "prebuild": "rimraf dist",
+    "prebuild": "rm -r dist || true",
```
thysultan, line 28: Does this word on windows?

heyheyhello: Depends if they have WSL. Not likely. I'll revert since Windows users are a consideration

package.json
```diff
@@ -36,14 +36,14 @@
...
     "esm": "3.2.25",
-    "mocha": "7.0.0",
-    "nyc": "15.0.0",
-    "rimraf": "3.0.2",
-    "rollup": "1.28.0",
-    "rollup-plugin-size": "0.2.1",
-    "rollup-plugin-terser": "5.1.3",
+    "mocha": "8.0.1",
+    "nyc": "15.1.0",
+    "rimraf": "^3.0.2",
```
thysultan, line 43: Pinned deps. ^3.0.2 vs 3.0.2

heyheyhello: It may have been less work to make that change yourself from GitHub 😉

How about we commit a package-lock.json if you're concerned about pinning dependencies?

thysultan:
CI job seems to be tripping(though tests pass) https://github.com/thysultan/stylis.js/pull/232/checks?check_run_id=887462067

heyheyhello:
Yeah I don't know why. Maybe you can kick it off again? Does your UI have a rebuild button or anything?

thysultan:
Yes, re-runned, it's related to the coveralls report:

```
Run coverallsapp/github-action@v1.0.1
Using lcov file: ./coverage/lcov.info
[error] "2020-07-20T04:52:18.840Z"  'error from getOptions'
##[error]Error: Command failed: git cat-file -p 8cf6dfc8db999543d3b7a4da8a6d1d35738be550
fatal: Not a valid object name 8cf6dfc8db999543d3b7a4da8a6d1d35738be550
```

I wonder if updating it to point to coverallsapp/github-action@master would fix it per https://github.com/coverallsapp/github-action#standard-example though that repo doesn't have a changelog to see if something changed to break it on that end.

Issue #233, opened 2020-07-16
.cjs does not work with React Native and/or Expo


efoken:
Since the latest version, which changed the file extension to `.cjs`, it does not work anymore with React Native. I'm getting the following error: 

```
Failed building JavaScript bundle.
While trying to resolve module `stylis` from file `/Users/.../node_modules/@emotion/cache/dist/cache.browser.cjs.js`, the package `/Users/.../node_modules/stylis/package.json` was successfully found. However, this package itself specifies a `main` module field that could not be resolved (`/Users/.../node_modules/stylis/dist/stylis.cjs`. Indeed, none of these files exist:

  * /Users/.../node_modules/stylis/dist/stylis.cjs(.native|.ios.expo.ts|.native.expo.ts|.expo.ts|.ios.expo.tsx|.native.expo.tsx|.expo.tsx|.ios.expo.js|.native.expo.js|.expo.js|.ios.expo.jsx|.native.expo.jsx|.expo.jsx|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.js|.native.js|.js|.ios.jsx|.native.jsx|.jsx|.ios.json|.native.json|.json|.ios.wasm|.native.wasm|.wasm)
  * /Users/.../node_modules/stylis/dist/stylis.cjs/index(.native|.ios.expo.ts|.native.expo.ts|.expo.ts|.ios.expo.tsx|.native.expo.tsx|.expo.tsx|.ios.expo.js|.native.expo.js|.expo.js|.ios.expo.jsx|.native.expo.jsx|.expo.jsx|.ios.ts|.native.ts|.ts|.ios.tsx|.native.tsx|.tsx|.ios.js|.native.js|.js|.ios.jsx|.native.jsx|.jsx|.ios.json|.native.json|.json|.ios.wasm|.native.wasm|.wasm)
```

Andarist:
It really seems like the Metro bundler/Expo problem. While we could "fix" this on our side this should definitely be fixed on theirs as .cjs will become more common in the ecosystem in the future.

sanpoChew:
there's an issue open on the metro repo https://github.com/facebook/metro/issues/535

hueter:
Unfortunately none of the [work-arounds](https://github.com/emotion-js/emotion/issues/1986#issuecomment-685952884) floating about are working for me on an Expo project, so this issue is blocking the upgrade to Emotion 11.

[...]

PR #238, opened 2020-09-04
Add react-native field to pkg.json

Andarist:
fixes https://github.com/thysultan/stylis.js/issues/233

It seems that the Metro bundler used by React Native doesn't handle either `.mjs` nor `.cjs` - which is somewhat incorrect because files with unknown extensions have always just been loaded as `.js` in node. Some more info can be found in my comment [here](https://github.com/emotion-js/emotion/issues/1986#issuecomment-685133928) as well.

While technically it's not our issue and theirs - it's unknown when this will get fixed on their side and there are React Native consumers confused with this as stylis is not working for them out of the box. I believe that it's best to just fix this on our side for now.

This probably could have also been fixed by reverting this commit: https://github.com/thysultan/stylis.js/commit/180b9829101cf0132c5671a36b9e2daf0a5e28e7 . Pick your poison 😉

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "type": "module",
   "main": "dist/stylis.cjs",
   "module": "dist/stylis.mjs",
+  "react-native": "./index.js",
   "exports": {
     "import": "./index.js",
     "require": "./dist/stylis.cjs"


Andarist:
@thysultan friendly 🏓

thysultan:
Published patch update to npm. Thanks for the ping.

PR #241, opened 2020-11-14
Make pkg.json#main a .js file

Andarist:
JavaScript ecosystem strikes again 🤦 Explained here: https://github.com/emotion-js/emotion/issues/2103#issuecomment-727207340

I will understand if you don't want to merge this but it would make my life easier if this could be released as people wouldn't be hurt by the ecosystem tooling quirks and I wouldn't have to explain this 😢

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "sideEffects": false,
   "type": "module",
-  "main": "dist/stylis.cjs",
+  "main": "dist/stylis.js",
   "module": "dist/stylis.mjs",
   "react-native": "./index.js",
   "exports": {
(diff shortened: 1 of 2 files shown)


thysultan:
Would the reason that we changed this in the first place surface as an issue again?

https://github.com/thysultan/stylis.js/pull/207#discussion_r409460723

Andarist:
> Would the reason that we changed this in the first place surface as an issue again?

Not as far as I can tell - `.cjs` has to be used because this is `type: "module"` package but it only has to be used within `pkg.json#exports` and this has been left untouched. This only affects `pkg.json#main` which should be ignored by any `pkg.json#exports`/`pkg.json#type`-aware tool.

thysultan:
In that case this looks good to merge.

Andarist:
Thanks! The only thing left to do is a new release then :)

PR #244, opened 2020-11-18
Fixed an issue with prefixer going into an infinite recursion

Andarist:
It's something that has been reported here: https://github.com/emotion-js/emotion/issues/2117

Gonna think about a potential fix for this later.

### DIFF
--- a/src/Prefixer.js
+++ b/src/Prefixer.js
@@ -67,16 +67,13 @@ export function prefix (value, length) {
 			if (strlen(value) - 1 - length > 6)
 				switch (charat(value, length + 1)) {
 					// (f)ill-available, (f)it-content
-					case 102:
-						if (charat(value, length + 3) === 108)
-							return replace(value, /(.+:)(.+)-([^]+)/, '$1' + WEBKIT + '$2-$3' + '$1' + MOZ + '$3') + value
+					case 102: length = charat(value, length + 3)
 					// (m)ax-content, (m)in-content
 					case 109:
-						return replace(value, /(.+:)(.+)-([^]+)/, '$1' + WEBKIT + '$2-$3' + '$1' + MOZ + '$2-$3') + value
-
+						return replace(value, /(.+:)(.+)-([^]+)/, '$1' + WEBKIT + '$2-$3' + '$1' + MOZ + (length == 108 ? '$3' : '$2-$3')) + value
 					// (s)tretch
 					case 115:
-						return prefix(replace(value, 'stretch', 'fill-available'), length) + value
+						return ~indexof(value, 'stretch') ? prefix(replace(value, 'stretch', 'fill-available'), length) + value : value
 				}
 			break
 		// position: sticky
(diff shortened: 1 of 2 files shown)


src/Prefixer.js
```diff
@@ -63,18 +63,19 @@ export function prefix (value, length) {
 		case 8116: case 7059: case 5753: case 5535:
 		case 5445: case 5701: case 4933: case 4677:
 		case 5533: case 5789: case 5021: case 4765:
+			var temp = value
```
Andarist, line 66: not ideal, but introducing a `temp` is the best I could figure out for this 🤷‍♂️

src/Prefixer.js
```diff
@@ -63,18 +63,19 @@ export function prefix (value, length) {
...
 					// (m)ax-content, (m)in-content
 					case 109:
 						return replace(value, /(.+:)(.+)-([^]+)/, '$1' + WEBKIT + '$2-$3' + '$1' + MOZ + '$2-$3') + value
-					// (f)ill-available
-					case 102:
-						return replace(value, /(.+:)(.+)-([^]+)/, '$1' + WEBKIT + '$2-$3' + '$1' + MOZ + '$3') + value
 					// (s)tretch
 					case 115:
-						return prefix(replace(value, 'stretch', 'fill-available'), length) + value
+						temp = (value = replace(value, 'stretch', 'fill-available')) + temp
```
thysultan, line 75:
```js
return indexof(value, 'stretch') > -1 ? ... : value
```

Andarist: the indexof approach doesnt work because of this `fit-content`:
https://github.com/thysultan/stylis.js/blob/9799734dcf2b04c1216818f8b74cba92b6de4f76/test/Prefixer.js#L130

this is actually missing from the comment above `case 102:`, should be:
```
// (f)ill-available, (f)it-content
```

thysultan: This should work

```js
// (min|max)?(width|height|inline-size|block-size)
case 8116: case 7059: case 5753: case 5535:
case 5445: case 5701: case 4933: case 4677:
case 5533: case 5789: case 5021: case 4765:
	// stretch, max-content, min-content, fill-available
	if (strlen(value) - 1 - length > 6)
		switch (charat(value, length + 1)) {
			// (f)ill-available, (f)it-content
			case 102: length = charat(value, length + 3)
			// (m)ax-content, (m)in-content
			case 109:
				return replace(value, /(.+:)(.+)-([^]+)/, '$1' + WEBKIT + '$2-$3' + '$1' + MOZ + (length == 108 ? '$3' : '$2-$3')) + value
			// (s)tretch
			case 115:
				return ~indexof(value, 'stretch') ? prefix(replace(value, 'stretch', 'fill-available'), length) + value : value
		}
	break
```

Andarist: Ok, went with that one

thysultan:
LGTM

src/Prefixer.js
thysultan: Did you see the `case 102: length = charat(value, length + 3)` amendment?

Andarist: Oh, no - I've focused on the `indexof` part in a hurry. This is supposed only to compress the code, right? Either way - I've pushed out this change right now.

Andarist:
Given that the final change is really proposed by you - I'm merging this right away.

thysultan:
Published v4.0.4 with these last 2 PR's. 

Thanks

Issue #247, opened 2020-12-01
ERR_REQUIRE_ESM error from requiring stylis


rcreagha:
I have a web app that uses emotion, which depends on stylis. While updating some dependencies I started getting the error seen below. The issue seems to be specific to 4.0.4, when I use 4.0.3 everything works fine.

```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: myApp/node_modules/stylis/dist/stylis.js
    require() of ES modules is not supported.
    require() of myApp/node_modules/stylis/dist/stylis.js from myApp/node_modules/@emotion/babel-plugin/dist/emotion-babel-plugin.cjs.dev.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
    Instead rename stylis.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from myApp/node_modules/stylis/package.json.

      at Object.<anonymous> (node_modules/@emotion/babel-plugin/dist/emotion-babel-plugin.cjs.dev.js:16:14)
```

Andarist:
This is weird - you should be loading this:
https://github.com/thysultan/stylis.js/blob/99afc806acf31a2b9a82d2d5439bba819e35eca6/package.json#L17
over this:
https://github.com/thysultan/stylis.js/blob/99afc806acf31a2b9a82d2d5439bba819e35eca6/package.json#L12

Could you prepare a runnable repro case? Sharing your node version etc could be helpful too.

eidietrich:
Also seeing this error today, with Node v12.16.1. Reverting to 4.0.3 also fixed the problem for me.

Andarist:
@eidietrich could you prepare a repro case on which I could take a look at?

thysultan:
@Andarist might this be relevant to https://github.com/thysultan/stylis.js/pull/241#issuecomment-727255242

belgattitude:
I can confirm the latest 4.0.4 packaging broke our app.

Context: it's a yarn workspace monorepo with nextjs + emotion (@emotion/babel-plugin..)

For those who are lucky enough to not rely on previous versions (like ^3.0 - a `yarn why stylis` will help to see) a quick hack is to force a module resolution in package.json

```
  "resolutions": {
    "stylis": "4.0.3"
  }
```

I'm trying to figure out a workaround, but crossing my fingers to have a fixed release asap. 

[...]

PR #248, opened 2020-12-02
Create commonjs package scope for the package.json#main file

Andarist:
fixes https://github.com/thysultan/stylis.js/issues/247

I totally forgot that this thing was an issue - I have totally forgotten that there are versions of node that:
- have experimental (!) but unflagged (!) ESM support
- do not have support for conditional exports (at least not an unflagged one)

We already had this problem which got fixed here: https://github.com/thysultan/stylis.js/pull/228 🤦 

I've also attempted this exact thing before in https://github.com/thysultan/stylis.js/commit/eb58ad5549af328a361464f86864eb75a4c73a3b which was part of https://github.com/thysultan/stylis.js/pull/207 but we have settled on the alternative solution there so this hasn't ever been merged to the master branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,12 +9,12 @@
   "bugs": "https://github.com/thysultan/stylis.js/issues",
   "sideEffects": false,
   "type": "module",
-  "main": "dist/stylis.js",
+  "main": "dist/umd/stylis.js",
   "module": "dist/stylis.mjs",
   "react-native": "./index.js",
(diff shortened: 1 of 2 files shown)


belgattitude:
Thanks millions @Andarist ! What a speed :1st_place_medal:

Andarist:
Gif with 'This is how we roll' phrase

PR #249, opened 2020-12-02
Fix package.json exports

Graphmaxer:
@thysultan This patch is very important because right now stylis is not usable. Node 15 tells me `Invalid package config <my_project>\node_modules\stylis\package.json. "exports" cannot contain some keys starting with '.' and some not. The exports object must either be an object of package subpath keys or an object of main entry condition name keys only.` Sorry about this mistake I didn't about this. This should be fine after merging this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,10 @@
   "module": "dist/stylis.mjs",
   "react-native": "./index.js",
   "exports": {
-    "import": "./index.js",
-    "require": "./dist/umd/stylis.js",
+    "." : {
+      "import": "./index.js",
+      "require": "./dist/umd/stylis.js",
+    },
     "./package.json": "./package.json"
   },
   "files": [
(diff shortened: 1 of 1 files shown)


Andarist (approved):

Graphmaxer:
Oops i missed a comma

package.json
```diff
@@ -13,8 +13,10 @@
   "module": "dist/stylis.mjs",
   "react-native": "./index.js",
   "exports": {
-    "import": "./index.js",
```
Andarist, line 16: makes sense - import/require are "conditions" and `./package.json` is not

thysultan:
Included in 4.0.6

Graphmaxer:
Thanks a lot @thysultan, @Andarist and @ljharb for the clarification. My warning is now gone ! :)

PR #251, opened 2020-12-08
Remove empty declarations

Andarist:
Has been reported here: https://github.com/emotion-js/emotion/issues/2162

Apparently, it's a behavioral change from Stylis 3. I believe that it would be great to handle this case but it's up for discussion

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -69,7 +69,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 					case 0: case 125: scanning = 0
 					// ;
 					case 59 + offset:
-						if (property > 0)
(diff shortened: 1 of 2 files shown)


src/Parser.js
```diff
@@ -69,7 +69,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 					case 0: case 125: scanning = 0
 					// ;
 					case 59 + offset:
-						if (property > 0)
+						if (property > 0 && (strlen(characters) - length))
```
Andarist, line 72: the idea behind the fix is that it checks if anything has been consumed for the value, if the diff is 0 then no - nothing has been consumed

PR #258, opened 2021-03-20
Fix position:sticky; prefix for Safari browser

### LINKED ISSUE #257: Position sticky is incorrectly prefixed
**Current behavior:**

```
prefix(`position:sticky;`, 8)
```
this code produces incorrect CSS:
```
-webkit-position:sticky;position:sticky;
```

**To reproduce:**

https://codesandbox.io/s/stilis-sticky-webkit-prefix-f6bro?file=/src/App.js:210-239

**Expected behavior:**

Generated CSS should look like this
```
position: -webkit-sticky; /* Safari */
position: sticky;
```
Source:
https://www.w3schools.com/howto/howto_css_sticky_element.asp

**Environment information:**

- `stylis` version: 4.0.7

isBatak:
Fix #257

Source: https://www.w3schools.com/howto/howto_css_sticky_element.asp

### DIFF
--- a/src/Prefixer.js
+++ b/src/Prefixer.js
@@ -86,7 +86,7 @@ export function prefix (value, length) {
 			switch (charat(value, strlen(value) - 3 - (~indexof(value, '!important') && 10))) {
 				// stic(k)y
 				case 107:
-					return replace(value, 'sticky', WEBKIT + 'sticky') + value
+					return replace(value, ':', ':' + WEBKIT) + value
 				// inline-b(o)x
 				case 111:
 					return replace(value, value, WEBKIT + value) + value
--- a/src/Prefixer.js
+++ b/src/Prefixer.js
@@ -84,8 +84,11 @@ export function prefix (value, length) {
 		// display: (flex|inline-flex|inline-box)
 		case 6444:
 			switch (charat(value, strlen(value) - 3 - (~indexof(value, '!important') && 10))) {
-				// stic(k)y, inline-b(o)x
-				case 107: case 111:
+				// stic(k)y
+				case 107:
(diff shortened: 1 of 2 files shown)


src/Prefixer.js
```diff
@@ -84,8 +84,11 @@ export function prefix (value, length) {
 		// display: (flex|inline-flex|inline-box)
 		case 6444:
 			switch (charat(value, strlen(value) - 3 - (~indexof(value, '!important') && 10))) {
-				// stic(k)y, inline-b(o)x
-				case 107: case 111:
+				// stic(k)y
+				case 107:
+					return replace(value, 'sticky', WEBKIT + 'sticky') + value
```
Andarist, line 89: Could be written as:
```suggestion
					return replace(value, ':', ':' + WEBKIT) + value
```

thysultan:
In that case `inline-box` should also have the same issue.

Andarist:
@thysultan what is inline-box anyway? can't really find quickly anything about it (apart from maybe just this one: https://caniuse.com/mdn-css_properties_display_xul_box_values )

thysultan:
@Andarist old flexbox version of `display: inline-flex`.

src/Prefixer.js
thysultan: Yes this would handle both.

Andarist:
@thysultan shouod we care about prefixing non-standard properties to non-standard properties though? 😅

thysultan:
Yeah i guess not, shouldn't be there, the one below it handles `inline-flex` to `inline-box` already.

Andarist:
@thysultan K, I've merged this one in and just pushed out to master a separate commit dropping `inline-box` prefixing. I hope that's OK

PR #260, opened 2021-04-04
fix / Escape chars in identifiers

richard-jp-leguen:
While working with emotion-js/styled, my team discovered that stylis is unable to parse CSS selectors which include escaped "special" characters. So a test like this:

```
  // add to test/Parser.js
  test.only('escaped chars in selector identifiers', () => {
    expect(
      stylis(`
        &.B\\&W{color:red;}
     `)
    ).to.equal(`.user.B\\&W{color:red;}`)
  });
```

Fails like this:

```
  1) Parser
       escaped chars in selector identifiers:

      AssertionError: expected '.user.B\\.userW{color:red;}' to equal '.user.B\\&W{color:red;}'
      + expected - actual

      -.user.B\.userW{color:red;}
      +.user.B\&W{color:red;}
      
      at Context.<anonymous> (test/Parser.js:140:10)
      at processImmediate (node:internal/timers:464:21)
```

Other special characters fail more spectacularly. Using `@` for example:

```
  test.only('escaped chars in selector identifiers', () => {
    expect(
      stylis(`
        &.B\\@W{color:red;}
     `)
    ).to.equal(`.user.B\\@W{color:red;}`)
  });
```

... yields the following result, with form feeds in the output:

```
1) Parser
       escaped chars in selector identifiers:

      AssertionError: expected '&\f.B\\@W{color:red;}' to equal '.user.B\\@W{color:red;}'
      + expected - actual

      -&
        .B\@W{color:red;}
      +.user.B\@W{color:red;}
      
      at Context.<anonymous> (test/Parser.js:140:10)
      at processImmediate (node:internal/timers:464:21)
```

-------

Various versions of the CSS recommendations mention that escape characters can be used.

From https://www.w3.org/TR/CSS2/syndata.html#characters
>  Identifiers can also contain escaped characters and any ISO 10646 character as a numeric code (see next item). For instance, the identifier "B&W?" may be written as "B\&W\?" or "B\26 W\3F".

From https://www.w3.org/TR/css-syntax-3/#escaping (non-normative)
> Any Unicode code point can be included in an identifier or quoted string by escaping it. CSS escape sequences start with a backslash (\), and continue with:
> * Any Unicode code point that is not a hex digits or a newline. The escape sequence is replaced by that code point.
> * Or one to six hex digits, followed by an optional whitespace. The escape sequence is replaced by the Unicode code point whose value is given by the hexadecimal digits. This optional whitespace allow hexadecimal escape sequences to be followed by "real" hex digits.

So in this PR I'm attempting to add some support for those behaviors to `stylis.js`

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -99,6 +99,18 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 				if (from(peek()).match(/[^a-f0-9]/i)) {
 					characters += '\\'
 					characters += from(next())
+				} else {
+					characters += '\\';
+					[1,2,3,4,5,6].some(() => {
+						if((from(peek()).match(/[a-f0-9]/i))) {
+							characters += from(next())
+							return false
+						}
+						return true
+					})
+					if((from(peek()).match(/\s/))) {
+						characters += from(next())
+					}
 				}
 				break
 			// :
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -94,6 +94,13 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 
 				index = offset = property = 0, variable = ampersand = 1, type = characters = '', length = pseudo
 				break
+			// \
+			case 92:
+				if (from(peek()).match(/[^a-f0-9]/i)) {
+					characters += '\\'
+					characters += from(next())
+				}
+				break
 			// :
 			case 58:
 				length = 1 + strlen(characters), property = previous
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -1,6 +1,6 @@
 import {COMMENT, RULESET, DECLARATION} from './Enum.js'
 import {abs, trim, from, sizeof, strlen, substr, append, replace} from './Utility.js'
-import {node, char, prev, next, peek, caret, alloc, dealloc, delimit, whitespace, identifier, commenter} from './Tokenizer.js'
+import {node, char, prev, next, peek, caret, alloc, dealloc, delimit, whitespace, escaping, identifier, commenter} from './Tokenizer.js'
 
 /**
  * @param {string} value
@@ -49,6 +49,10 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 			case 9: case 10: case 13: case 32:
 				characters += whitespace(previous)
 				break
+			// \
+			case 92:
+				characters += escaping(caret() - 1, 7)
+				continue
 			// /
 			case 47:
 				switch (peek()) {
@@ -94,36 +98,15 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 
 				index = offset = property = 0, variable = ampersand = 1, type = characters = '', length = pseudo
 				break
-			// \
-			case 92:
-				if (from(peek()).match(/[^a-f0-9]/i)) {
-					characters += '\\'
-					characters += from(next())
-				} else {
-					characters += '\\';
-					[1,2,3,4,5,6].some(() => {
-						if((from(peek()).match(/[a-f0-9]/i))) {
-							characters += from(next())
-							return false
-						}
-						return true
-					})
-					if((from(peek()).match(/\s/))) {
-						characters += from(next())
-					}
-				}
-				break
 			// :
 			case 58:
 				length = 1 + strlen(characters), property = previous
 			default:
-				if (variable < 1) {
-					if (character == 123) {
+				if (variable < 1)
+					if (character == 123)
 						--variable
-					} else if (character == 125 && variable++ == 0 && prev() == 125) {
+					else if (character == 125 && variable++ == 0 && prev() == 125)
 						continue
-					}
-				}
 
 				switch (characters += from(character), character * variable) {
 					// &
--- a/test/Parser.js
+++ b/test/Parser.js
@@ -138,8 +138,16 @@ describe('Parser', () => {
         &.B\\&W{color:red;}
         &.\\@example\\.com{color:blue;}
         &.owner\\/founder{color:green;}
+        &.discount\\%  {color:purple;}
      `)
-    ).to.equal(`.user.B\\&W{color:red;}.user.\\@example\\.com{color:blue;}.user.owner\\/founder{color:green;}`)
+    ).to.equal([
+      '.user.B\\&W{color:red;}',
+      '.user.\\@example\\.com{color:blue;}',
+      '.user.owner\\/founder{color:green;}',
+      // while doubel spaces after escaped hex codes need to be preserved,
+      // after an escaped character / code point it need not be preserved 
+      '.user.discount\\%{color:purple;}'
+    ].join(''))
   });
 
   test('escaped hex codes in selector identifiers', () => {
--- a/test/Parser.js
+++ b/test/Parser.js
@@ -132,6 +132,16 @@ describe('Parser', () => {
     ).to.equal(`.user [href="https://css-tricks.com?a=1&b=2"]{color:red;}`)
   })
 
+  test('escape chars in selector identifiers', () => {
+    expect(
+      stylis(`
+        &.B\\&W{color:red;}
+        &.\\@example\\.com{color:blue;}
+        &.owner\\/founder{color:green;}
+     `)
+    ).to.equal(`.user.B\\&W{color:red;}.user.\\@example\\.com{color:blue;}.user.owner\\/founder{color:green;}`)
+  });
+
   test('comments', () => {
     expect(
       stylis(`
--- a/test/Parser.js
+++ b/test/Parser.js
@@ -144,7 +144,7 @@ describe('Parser', () => {
       '.user.B\\&W{color:red;}',
       '.user.\\@example\\.com{color:blue;}',
       '.user.owner\\/founder{color:green;}',
-      // while doubel spaces after escaped hex codes need to be preserved,
+      // while double spaces after escaped hex codes need to be preserved,
       // after an escaped character / code point it need not be preserved 
       '.user.discount\\%{color:purple;}'
     ].join(''))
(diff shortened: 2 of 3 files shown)


src/Parser.js
```diff
@@ -94,6 +94,25 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 
 				index = offset = property = 0, variable = ampersand = 1, type = characters = '', length = pseudo
 				break
+			// \
+			case 92:
```
richard-jp-leguen, line 98: I'm a little worried about the fact that there's nothing here to distinguish between an escape sequence in an "identifier" as opposed to somewhere else, but I can't think of a test case where that's going to be a problem.

src/Parser.js
```diff
@@ -94,6 +94,25 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
...
+				if (from(peek()).match(/[^a-f0-9]/i)) {
+					characters += '\\'
+					characters += from(next())
+				} else {
+					characters += '\\';
+					[1,2,3,4,5,6].some(() => {
```
richard-jp-leguen, line 104: I hesitated to add a `for(var i` loop, as all the variables are hoisted and use the `var` keyword - I didn't want to pollute that list with a scarcely-used `i`.

test/Parser.js
```diff
@@ -132,6 +132,36 @@ describe('Parser', () => {
...
+        &.B\\26 W{color:blue;}
+        &.endsWith\\0000A9  a.childNode{color:green;}
+        &.endsWith\\AE  a.childNode{color:yellow;}
+     `)
+    ).to.equal([
+      '.user.B\\26W{color:red;}',
+      '.user.B\\000026W{color:green;}',
+      '.user.B\\26 W{color:blue;}',
+      // next rules are important because the hex digits terminating space
+      // is not the same as a combinator space
```
richard-jp-leguen, line 159: Demonstration that the double-space is important: https://jsfiddle.net/6gsr7qv4/

src/Parser.js
```diff
@@ -94,6 +94,25 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 
 				index = offset = property = 0, variable = ampersand = 1, type = characters = '', length = pseudo
 				break
+			// \
+			case 92:
+				if (from(peek()).match(/[^a-f0-9]/i)) {
```
thysultan, line 99: This looks overly involved, why not just `next()` the `continue` i.e

```js
case 92: next()
    continue
```

richard-jp-leguen: Because then the tests fail!

```
 2 failing

  1) Parser
       escaped chars in selector identifiers:

      AssertionError: expected '.user.B{color:red;}.user.xampleom{color:blue;}.user.ownerounder{color:green;}' to equal '.user.B\\&W{color:red;}.user.\\@example\\.com{color:blue;}.user.owner\\/founder{color:green;}'
      + expected - actual

      -.user.B{color:red;}.user.xampleom{color:blue;}.user.ownerounder{color:green;}
      +.user.B\&W{color:red;}.user.\@example\.com{color:blue;}.user.owner\/founder{color:green;}
      
      at Context.<anonymous> (test/Parser.js:142:10)
      at processImmediate (node:internal/timers:464:21)

  2) Parser
       escaped hex digits in selector identifiers:

      AssertionError: expected '.user.BW{color:red;}.user.B0026W{color:green;}.user.B W{color:blue;}.user.endsWith00A9 a.childNode{color:green;}.user.endsWith a.childNode{color:yellow;}' to equal '.user.B\\26W{color:red;}.user.B\\000026W{color:green;}.user.B\\26 W{color:blue;}.user.endsWith\\0000A9  a.childNode{color:green;}.user.endsWith\\AE  a.childNode{color:yellow;}'
      + expected - actual

      -.user.BW{color:red;}.user.B0026W{color:green;}.user.B W{color:blue;}.user.endsWith00A9 a.childNode{color:green;}.user.endsWith a.childNode{color:yellow;}
      +.user.B\26W{color:red;}.user.B\000026W{color:green;}.user.B\26 W{color:blue;}.user.endsWith\0000A9  a.childNode{color:green;}.user.endsWith\AE  a.childNode{color:yellow;}
      
      at Context.<anonymous> (test/Parser.js:154:10)
      at processImmediate (node:internal/timers:464:21)
```

The output still needs the `\` and the next char, so I assume we need a `characters += ` to appear somewhere.

And we sometimes need to preserve double spaces `  ` when they follow hex code sequences, but not when they follow escaped non-hex code characters.

thysultan: >so I assume we need a characters += to appear somewhere.

Yes we can concat the two characters that follow:

```js
case 92: characters += from(next()) + from(next())
    continue
```

richard-jp-leguen: At a glance - this skips the slash? Would it be more appropriate to do:

```
case 92: characters += from(character) + from(next())
    continue;
```

That still won't pass one of the tests - this solution does nothing for the escaped 6 character hex code followed by a double space? https://jsfiddle.net/8ouxLvdf/

Changing the code, the tests say:

```
1) Parser
       double spaces after escaped hex codes in selector identifiers:

      AssertionError: expected '.user.endsWith\\0000A9 a.childNode{color:green;}.user.endsWith\\AE a.childNode{color:yellow;}.user.Q\\000026A a.childNode{color:purple;}' to equal '.user.endsWith\\0000A9  a.childNode{color:green;}.user.endsWith\\AE  a.childNode{color:yellow;}.user.Q\\000026A a.childNode{color:purple;}'
      + expected - actual

      -.user.endsWith\0000A9 a.childNode{color:green;}.user.endsWith\AE a.childNode{color:yellow;}.user.Q\000026A a.childNode{color:purple;}
      +.user.endsWith\0000A9  a.childNode{color:green;}.user.endsWith\AE  a.childNode{color:yellow;}.user.Q\000026A a.childNode{color:purple;}
      
      at Context.<anonymous> (test/Parser.js:166:10)
      at processImmediate (node:internal/timers:464:21)
```

test/Parser.js
thysultan: Is `\000026A `valid unicode wouldn't it be `\26A `?

richard-jp-leguen: It's valid. From the recommendation I linked to in the PR description, https://www.w3.org/TR/css-syntax-3/#escaping

> EXAMPLE 3
> An identifier with the value "&B" could be written as `\26 B` or `\000026B`.

richard-jp-leguen: ... although the trailing `A` isn't part of the hex code, since the hex code is "one to six hex digits". In both what commented and what's in the recommendation's EXAMPLE 3, the hex code is `\26` (or `\000026`) and the `A` or `B` which follows is _not_ part of the hex code.

richard-jp-leguen: ... as a matter of fact, because `A` is a hex digit, `\000026A` and `\26A` are different - one is `&A` and the other is `ɪ`, because with neither leading `0`s nor a whitespace after the hex code, the `A` is treated as part of the hex code!

So:

* `Q\000026A` -> `Q&A`
* `Q\000026 A` -> `Q&A`
* `Q\26 A` -> `Q&A`
* `Q\26A` -> `Qɪ`

thysultan: So all unicode sequences need two spaces to separate them i.e `Q \000026 A` would need to be `Q \000026  A` with two spaces to yield `Q & A`?

richard-jp-leguen: I think github collapses spaces for inline code snippets, and I can't follow if you're trying to use a className of `Q & A` (which wouldn' t work because classNames are delimited by spaces in the HTML) or trying to use space as a descendant combinator.

If you had something like `id="Q & A"` you'd need to escape the spaces just the same as the ampersand, and since the escaping slash is not a hex digit you wouldn't need a double space.

If you wanted to have a space which is a combinator after the hex digits, yes you'd need a double space.

Examples: https://jsfiddle.net/sz67trap/

thysultan: No meant selecting `\000026 A` where a is another selector i.e `\000026 header` wouldn't that now treat it as `unicode` + `header` without a space between them?

richard-jp-leguen: Yes, this rule:

```
.Q\000026 header {
    color: red;
}
```

Targets this element:

```
<div class="Q&header">
```

To target this:

```
<div class="Q&">
    <header>
```

You would need a double space in the selector rule - one to terminate the hex code and one to use as a combinator:

```
.Q\000026  header {
    color: blue;
}
```

See also https://jsfiddle.net/4heovdau/

------

This is accounted for in the tests:
* a single space after 6-0r-fewer hex digits needs to be preserved: https://github.com/thysultan/stylis.js/pull/260/files#diff-4cafb040dcdb12664448ab294d801e96bb1378853e3a2a4337d280b1d3de7f89R158
* a double space after 6-or-fewer hex digits needs to be preserved: https://github.com/thysultan/stylis.js/pull/260/files#diff-4cafb040dcdb12664448ab294d801e96bb1378853e3a2a4337d280b1d3de7f89R170-R171
* a double space after a non-hex escape sequence is just a combinator and can be collapsed into a single whitespace char: https://github.com/thysultan/stylis.js/pull/260/files#diff-4cafb040dcdb12664448ab294d801e96bb1378853e3a2a4337d280b1d3de7f89R141
* a double space after the 7th hex digit is no longer a hex code terminating space but just a run of the mill combinator space which can be collapsed: https://github.com/thysultan/stylis.js/pull/260/files#diff-4cafb040dcdb12664448ab294d801e96bb1378853e3a2a4337d280b1d3de7f89R172

src/Parser.js
```diff
@@ -49,6 +49,10 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 			case 9: case 10: case 13: case 32:
 				characters += whitespace(previous)
 				break
+			// \
+			case 92:
+				characters += escaping(caret() - 1, 7)
```
richard-jp-leguen, line 54: Really like adding a function to `Tokenizer.js` for this.

Andarist (approved):

PR #273, opened 2021-09-11
Add support for ampersand in pseudo selectors

### LINKED ISSUE #272: Referencing & from within :not()
A bit of an odd case, but do you think this is behaving correctly?

```css
.foo {
    .bar:not(&) {
        color: red;
    }
}
```

I would have expected this to produce:

```css
.bar:not(.foo) {
    color: red;
}
```

Instead it seems to ignore the ampersand inside `:not()`:

Screen Shot 2021-09-09 at 2 38 46 PM

stonebk:
Fixes #272

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,11 +39,11 @@
     "release-patch": "npm version patch -m '%s'"
   },
   "devDependencies": {
-    "chai": "4.2.0",
+    "chai": "^4.3.4",
     "eslint": "6.8.0",
     "esm": "3.2.25",
-    "mocha": "7.0.0",
-    "nyc": "15.0.0",
+    "mocha": "^9.1.1",
+    "nyc": "^15.1.0",
     "rimraf": "3.0.2",
     "rollup": "1.28.0",
     "rollup-plugin-size": "0.2.1",
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -1,5 +1,5 @@
 import {COMMENT, RULESET, DECLARATION} from './Enum.js'
-import {abs, trim, from, sizeof, strlen, substr, append, replace, indexof} from './Utility.js'
+import {abs, trim, from, sizeof, strlen, substr, append, replace} from './Utility.js'
 import {node, char, prev, next, peek, caret, alloc, dealloc, delimit, whitespace, escaping, identifier, commenter, slice} from './Tokenizer.js'
 
 /**
@@ -39,25 +39,10 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 	var reference = rule
 	var characters = type
 	var colon = -1
+	var pseudoselector = 0
 
 	while (scanning)
 		switch (previous = character, character = next()) {
-			// (
-			case 40:
-				if (colon > 0) {
-					switch (slice(colon, caret() - 1)) {
-						case 'matches':
-						case 'is':
-						case 'not':
-						case 'any':
-						case 'where':
-						case 'has':
-							colon = delimit(character)
-							indexof(colon, '&') > 0 ? (ampersand = -1, characters += replace(colon, /&/g, '&\f')) : characters += colon
-							colon = -1
-							continue
-					}
-				}
 			// " ' [
 			case 34: case 39: case 91:
 				characters += delimit(character)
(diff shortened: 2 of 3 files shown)


stonebk:
Hey @Andarist, not sure if there is a cleaner way to do this -- this seems to be working though! Let me know if there are any changes I can make.

package.json
```diff
@@ -39,11 +39,11 @@
     "release-patch": "npm version patch -m '%s'"
   },
   "devDependencies": {
-    "chai": "4.2.0",
+    "chai": "^4.3.4",
     "eslint": "6.8.0",
     "esm": "3.2.25",
-    "mocha": "7.0.0",
-    "nyc": "15.0.0",
+    "mocha": "^9.1.1",
```
stonebk, line 45: I was running into errors when trying to run specs until I updated these outdated `devDependencies`.

stonebk:
The original implementation did not account for selectors with strings including ampersands, e.g. `:matches(&,[href="https://css-tricks.com?a=1&b=2"])`

I've added another test case and updated the implementation -- I think it's a bit cleaner now.

test/Parser.js
```diff
@@ -528,6 +528,78 @@ describe('Parser', () => {
...
+    expect(
+      stylis(`
+        h1:matches(.a,.b,:not(&)) {
+          display: none
+        }
+      `)
+    ).to.equal(`h1:matches(.a,.b,:not(.user)){display:none;}`)
+  })
+
+  test('pseudo-selector with & in a string', () => {
```
Andarist, line 593: excellent test case 👍

test/Parser.js
```diff
@@ -528,6 +528,78 @@ describe('Parser', () => {
...
+      `:is(.user){color:red;}`,
+      `:any(.user){color:red;}`,
+      `:where(.user){color:red;}`,
+      `:has(.user){color:red;}`
+    ].join(''))
+  })
+
+  test('nested pseudo-selectors', () => {
+    expect(
+      stylis(`
```
Andarist, line 565: I think it's the correct output here - the interesting bit is that Sass doesn't seem to handle this correctly

src/Parser.js
```diff
@@ -100,8 +102,28 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
...
+				// (
+				if (character === 40) {
+					if (colon > 0) {
+						switch (slice(colon, caret() - 1)) {
+							case 'matches':
+							case 'is':
+							case 'not':
+							case 'any':
+							case 'where':
+							case 'has':
```
Andarist, line 116: any particular reason why do we have to maintain the allowlist here? couldn't we just assume that this should always be treated as a pseudoselector?

stonebk: I think that's a great question. I think my primary reason for doing it this way was to minimize the surface area of my change. Perhaps it's not necessary -- would require some more thought (and likely some more test cases) about how other css rules would possibly be impacted.

stonebk: A quick test removing the allowlist causes issues with `calc` and `url`:

```
height:calc( 100vh - 1px );
background:url("data:image/png;base64,iVBORw0");
```

Andarist: Those are mostly used in the declarations - however, they can also appear in the selectors~. `calc` can be used within `@media` and `url` can be used together with `@document`. It seems to me though that for them to both appear in the selectors~ and be preceded by a colon is not likely at all - or am I missing something? I wonder if this could be used somehow here. 🤔

stonebk: I agree, the likelihood is low and we could probably simplify.

Right now my logic is basically:

```
if we saw a "(" preceded by ":matches", ":is", etc.
    continue parsing selector character by character
else
   skip parsing entire "(...)" block
```

What the logic could potentially be simplified to is:

```
if we saw a "(" and we're currently parsing a selector
    continue parsing selector character by character
else // we're parsing a declaration
   skip parsing entire "(...)" block
```

I think this is where I'm limited by my lack of familiarity with the code. Are we able to determine whether or not we are currently parsing a selector or a declaration?

test/Parser.js
stonebk: I think this is because `:is` and `:where` are broken:

```css
.foo {
  :not(&) {
    color: red;
  }

  :matches(&) {
    color: red;
  }

  :is(&) {
    color: red;
  }

  :any(&) {
    color: red;
  }

  :where(&) {
    color: red;
  }

  :has(&) {
    color: red;
  }
}
```

```css
:not(.foo) {
  color: red;
}
:matches(.foo) {
  color: red;
}
.foo :is(&) {
  color: red;
}
:any(.foo) {
  color: red;
}
.foo :where(&) {
  color: red;
}
:has(.foo) {
  color: red;
}
```

If I remove those, it looks like works:

```css
.foo {
    :not(&) {
        :matches(&) {
            :any(&) {
                :has(&) {
                    color: red;
                }
            }
        }
    }
}
```

```css
:has(:any(:matches(:not(.foo)))) {
  color: red;
}
```

(using https://www.sassmeister.com/)

src/Parser.js
Andarist: any thoughts on this @thysultan ?

thysultan:
Thanks, the latest patch 4.0.11 has this functionality now.

Issue #101, opened 2018-03-23
disabled cascade and specificity


giuseppeg:
We realized that disabling cascade and scoping every selector breaks specificity eg.

Say we have:

```css
div > div > span // specificity [0,0,0,3]
.clickTest // specificity [0,0,1,0]
```

When each selector is scoped the specificity becomes:

```
[0,0,3,3]
[0,0,2,0]
```

Now the first selector is more specific than the second - it wasn't the case with the original selectors.

To solve this issue I think that we could just scope the last selector:

```css
div > div > span.jsx-123 // specificity [0,0,1,3]
.clickTest.jsx-123 // specificity [0,0,2,0] <- this still wins
```
@thysultan would it be hard to make this change and maybe release a major version?

thysultan:
I think the current heuristics was the hardest part, that said would this change still isolate(in the broader sense) the cascade in the same fashion that the current heuristics intended it to?

giuseppeg:
we'd need to do some thinking and try to isolate the edge cases, eg. mixed scoped and global styles. But I think that most of the time this should work just fine.

thysultan:
>eg. mixed scoped and global styles. 

Do you have examples of these edge cases?

giuseppeg:
Yep there is an edge case with `:global` see the first example:
https://jsfiddle.net/aeznqkyw/

giuseppeg:
so when the last selector is `:global` we should scope the last non-global selector

giuseppeg:
I just checked and Vue uses the same technique (scoping the last selector). As I said above we also need to support the `:global` special case. How hard would it be to implement this way of scoping?

giuseppeg:
Ok here's another edge case https://jsfiddle.net/aeznqkyw/10/ see the _nested edge case_ basically if we only scope the innermost selector other components can cause the selector to match anyway. 

The solution I guess is to always scope the innermost and outermost selectors. To not break specificity then we need to add the scoped class twice when the selector is single eg. `div.jsx-123.jsx-123` https://jsfiddle.net/aeznqkyw/14/

thysultan:
Have i missed out anything in the following list:

```scss
// leading :global 
:global(div) div div {}
div div.ns div.ns {}

// trailing :global 
div div :global(div) {}
div.ns div.ns div {}

// inner :global
div :global(div) div span {}
div.ns div div span.ns {}

// single + leading :global
:global(span) div {}
span div.ns.ns {}

// single + trailing :global
[...]

PR #282, opened 2022-01-09
Update middleware namespace function and test

ywroh:


thysultan:
See https://github.com/thysultan/stylis.js/issues/101 for why we do this. Not a bug.

Issue #119, opened 2018-09-06
autoprefixer for grid in IE11?


jonasbjorkberg:
A year has gone by since this was last discussed, (https://github.com/thysultan/stylis.js/issues/51).
Since then Autoprefixer has implemented a decent support for css grid in IE11. See for example https://css-tricks.com/css-grid-in-ie-css-grid-and-the-new-autoprefixer/.
It would have been fantastic if stylis had the same support.

izakfilmalter:
Ya this is causing a lot of headaches for my dev team atm.

smeijer:
Would be awesome indeed. The fact that the IE grid is so different from any other implementation; prevents me from using it at all.

As long as IE is around, I stick to `Flex` *(which also has its share of issues (`-ms-flex-negative`))*.

Auto fixing grids would be **AWESOME**!

fredconv:
Hi 

Any news about this one ? That would be **awesome**.
I'm looking forward to be able to use grid in all our components without any extra support for IE :) 

👍 

Thanks

thysultan:
Open to review PR's on this if it doesn't bloat the bundle size a lot. Preferably on the `rewrite` branch where the prefixer is tree-shake-able.

bzalasky:
@thysultan I'd like to take a crack at this if it's still open. We're using Emotion at Lookout, which depends on Stylis, and would like to adopt CSS Grid, but need to continue supporting IE 11 for the time being.

thysultan:
@bzalasky Yes, feel free. Ping me if you have any questions.

bzalasky:
Looking into it, it seems Emotion uses a Stylis fork that’s “been modified slightly to make it smaller.” Though it looks like all they’re doing is running it through Google Closure Compiler and Prettier, so it seems any changes here could potentially land there as well… though not sure how the rewrite branch fits into this then. Does the rewrite branch include breaking changes? Is it at parity with master?

thysultan:
Yes the rewrite is major version bump so it might not make it into their release schedule. 

>Is it at parity with master?

It's not intended to maintain API parity with the current version.

fredconv:
Just to know if there was any update on this part ? :) 

thanks

JeroenReumkens:
@thysultan Could you point me in the right direction on how I could start adding this? I'd love to try to fix this, but at the moment don't have any idea where to start. I think this is a really important feature that is right now missing, since grid will is used more and more.

TylerBarnes:
[...]

PR #285, opened 2022-02-24
feat(#119): add grid prefix support

SukkaW:
When finished, this should close https://github.com/thysultan/stylis/issues/119 once and for all.

Note: ~~It is still WIP~~ Although the feature is implemented, it **hasn't** been optimized **AT ALL**, so **suggestions, optimizations, discussions are welcome**!


SukkaW:
TODO:

- [x] display: grid / inline-grid
- [ ] align-self / justify-self: strech / end / center
- [ ] grid-template-columns
- [ ] grid-template-rows
- [ ] grid-template-columns
- [ ] grid-row-start: numerical
- [ ] grid-row-end: numerical
- [ ] grid-row-start: non-numerical
- [ ] grid-row-end: non-numerical
- [ ] grid-column: expand shorthand syntax
- [ ] grid-row: expand shorthand syntax

SukkaW:
@thysultan @Andarist 

The PR is ready for review. And here is a few things:

- Prefixing `grid-row-start` and `grid-row-end` requires the knowledge of both declarations, so it is impossible for `stylis` stateless *per declaration* processor to prefix them.
- I just write the code to make it pass the test cases. It is **highly un-optimized**, introducing tons of performance overhead and unnecessary byte to the final dist build. Feel free to add your ideas and suggestions!

src/Prefixer.js
```diff
@@ -34,7 +34,12 @@ export function prefix (value, length) {
 			return WEBKIT + value + replace(value, /(\w+).+(:[^]+)/, WEBKIT + 'box-$1$2' + MS + 'flex-$1$2') + value
 		// align-self
 		case 5443:
-			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/, '') + value
+			// center, end, start, stretch
+			return (
+				~indexof(['stretch', 'center', 'end', 'start'], substr(value, length + 1, sizeof(value) - 1))
```
thysultan, line 39: TODO: convert to regex, not Array.indexOf.

src/Prefixer.js
```diff
@@ -59,6 +64,17 @@ export function prefix (value, length) {
 		// justify-content
 		case 4968:
 			return replace(replace(value, /(.+:)(flex-)?(.*)/, WEBKIT + 'box-pack:$3' + MS + 'flex-pack:$3'), /s.+-b[^;]+/, 'justify') + WEBKIT + value + value
+		// justify-self
+		case 4200:
+			// center, end, start, stretch
+			if (
+				~indexof(['stretch', 'center', 'end', 'start'], substr(value, length + 1, sizeof(value) - 1))
```
thysultan, line 71: TODO: convert to regex, not Array.indexOf.

src/Prefixer.js
```diff
@@ -96,6 +120,9 @@ export function prefix (value, length) {
 				// (inline-)?fl(e)x
 				case 101:
 					return replace(value, /(.+:)([^;!]+)(;|!.+)?/, '$1' + WEBKIT + (charat(value, 14) === 45 ? 'inline-' : '') + 'box$3' + '$1' + WEBKIT + '$2$3' + '$1' + MS + '$2box$3') + value
+				// (inline-)?grid
+				case 105:
+					return replace(value, ':', ':' + MS) + value
```
thysultan, line 125: Looks good 👍

src/Prefixer.js
```diff
@@ -82,12 +98,20 @@ export function prefix (value, length) {
 						return ~indexof(value, 'stretch') ? prefix(replace(value, 'stretch', 'fill-available'), length) + value : value
 				}
 			break
+		// grid-(column|row)
+		case 5152: case 5920:
+			// simple position value
+			if (match(substr(value, length + 1, sizeof(value) - 1), /^\d+$/)) {
+				return MS + value + value
+			} else if (match(substr(value, length + 1, sizeof(value) - 1), /(\d)+\s*\/( ?span)?\s*(\d)+/)) {
+				return replace(substr(value, length + 1, sizeof(value) - 1), /(\d)+\s*\/( ?span)?\s*(\d)+/, (_, $1, $span, $3) => MS + substr(value, 0, length + 1) + $1 + ';' + MS + substr(value, 0, length) + '-span:' + ($span ? $3 : +$3 - +$1) + ';') + value
```
thysultan, line 107: Will thoroughly review tomorrow evening.

thysultan:
> * Prefixing `grid-row-start` and `grid-row-end` requires the knowledge of both declarations

Elaborate?

thysultan:
> * so it is impossible for `stylis` stateless _per declaration_ processor to prefix them

The prefixer middleware, which has access to the complete state of the css token graph does not have this limitation, the `prefix` function in `Prefixer.js` is called by this middleware: https://github.com/thysultan/stylis/blob/9e61e05a04060ad397c951877dee311297825568/src/Middleware.js#L42

SukkaW:
> Elaborate?

If `grid-row-start` and `grid-row-end` should be prefixed, here is how:

- If only `grid-row-start` exists:
  - Prefix with `-ms-grid-row`. So `grid-row-start:3;` becomes `-ms-grid-row:3;-ms-grid-row:3`.
- If only `grid-row-end` exists:
  - Prefix with `-ms-grid-row-span`. So `grid-row-end:3;` becomes `-ms-grid-row-span:3;grid-row-end:3;`, and `grid-row-end:span 3;` becomes `-ms-grid-row-span:3;grid-row-end:span 3;`
- If both of them exists:
  - Prefix `grid-row-start` with `-ms-grid-row`
  - Add `-ms-grid-row-span` with a value of `grid-row-end - grid-row-start`
  - So `grid-row-start:3;grid-row-end:5;` becomes `-ms-grid-row: 3;grid-row-start:3;-ms-grid-row-span:2;grid-row-end:5;`

The same rule applies to `column`.

I mainly port the features from @robinweser's [inline-style-prefixer](https://github.com/robinweser/inline-style-prefixer) (He recently switch [Fela](https://fela.js.org)'s prefixer from his own `inline-style-prefixer` to `stylis`. Porting the feature helps to keep consistent behavior).

src/Prefixer.js
SukkaW: Fixed in https://github.com/thysultan/stylis/pull/285/commits/817ff44b0bbc87f950e25b9e35f72d7531f2ee81

src/Prefixer.js
SukkaW: Fixed in https://github.com/thysultan/stylis/pull/285/commits/817ff44b0bbc87f950e25b9e35f72d7531f2ee81

Andarist:
One concern that I have is that it might potentially increase the output quite a bit for existing users. I wonder if this could be split into 2 plugins somehow? Or should I potentially get ready for copying over the older version of the prefixed into Emotion's codebase? 🤔

robinweser:
Chiming in since I was mentioned in this thread:
Maybe it'd be worth exploring a way to build your own prefixer from browser data or something?
Over at [inline-style-prefixer](https://github.com/robinweser/inline-style-prefixer), we export a `generateData` and `createPrefixer` function which one can use to create your very own one given a list of browser versions. See https://inline-style-prefixer.js.org/docs/guides/CustomPrefixer.html

SukkaW:
> One concern that I have is that it might potentially increase the output quite a bit for existing users. I wonder if this could be split into 2 plugins somehow? Or should I potentially get ready for copying over the older version of the prefixed into Emotion's codebase? 🤔

That's why I said my initial changes are `highly unoptimized`. And I believe it is possible to reduce the impact down to below 2 KiB or even below 1 KiB, which should be acceptable.

Andarist:
> And currently, the after-gzipped size is only increased by about 200 bytes. That doesn't sound so bad, right?

I'm not that concerned about the increase of the Stylis' size - but rather about the increased size of the prefixed output. My concern is about the generated CSS.

SukkaW:
> I'm not that concerned about the increase of the Stylis' size - but rather about the increased size of the prefixed output. My concern is about the generated CSS.

Well, after the change, only those grid-related declarations will contribute to the increase of the final CSS output. So it really depends on how much grid do the developers use in their project currently: if they use none, then the prefixed output will not be affected at all.

So I've done some naive tests here:

- Go to some websites
- Grab a dump of their entire CSS
- Put the CSS into PostCSS's autoprefixer, with the `browserlist` set to `since 2022, IE 10` to make sure only IE flexbox and IE grid prefixes get applied.
- Compare the size change.

I've checked [dev.to](https://dev.to) (their source code can be found [here](https://github.com/forem/forem)), a website whose layout is heavily based on the CSS grid. And their CSS size before and after autoprefixer with `since 2022, IE 10` are `444KiB` vs `449KiB`.

src/Prefixer.js
```diff
@@ -34,7 +34,14 @@ export function prefix (value, length) {
 			return WEBKIT + value + replace(value, /(\w+).+(:[^]+)/, WEBKIT + 'box-$1$2' + MS + 'flex-$1$2') + value
 		// align-self
 		case 5443:
-			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/, '') + value
+			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/, '')
+				+ (
+					// center, end, start, stretch
+					match(substr(value, length + 1, sizeof(value) - 1), /^(center|end|start|stretch)$/)
```
thysultan, line 40: flexbox supports plain `center|end|start|stretch` wouldn't this now add grid properties to what was intended as flexbox `align-self` given that `center|end|start|stretch` are universal properties for both flexbox and grid?

SukkaW: Well, at least it is also how autoprefixer handle `align-self`:

```css
.example {
    align-self: stretch
}
```

will become

```css
/*
* Prefixed by https://autoprefixer.github.io
* PostCSS: v8.3.6,
* Autoprefixer: v10.3.1
* Browsers: since 2010
*/

.example {
    -webkit-align-self: stretch;
        -ms-flex-item-align: stretch;
                -ms-grid-row-align: stretch;
            align-self: stretch
}
```

src/Prefixer.js
```diff
@@ -59,6 +66,15 @@ export function prefix (value, length) {
 		// justify-content
 		case 4968:
 			return replace(replace(value, /(.+:)(flex-)?(.*)/, WEBKIT + 'box-pack:$3' + MS + 'flex-pack:$3'), /s.+-b[^;]+/, 'justify') + WEBKIT + value + value
+		// justify-self
+		case 4200:
+			// center, end, start, stretch
+			if (
+				match(substr(value, length + 1, sizeof(value) - 1), /^(center|end|start|stretch)$/)
```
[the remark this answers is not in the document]
SukkaW: > Why not just match on the value i.e `match(value, /(cen|end|sta|str).*/)` ?

The value must be matched exactly, preventing `flex-(start|end)`, `self-(start|end)` from being prefixed.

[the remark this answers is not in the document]
thysultan: and the matching on value instead of substr(value, length + 1, sizeof(value - 1)) part, you're extracting the prop from `key:prop;` but since we are sure that the key cannot possibly match the list of options we don't need to exclude the key from the match? That is `center|end|start|stretch` cannot possibly be in `justify-self:`.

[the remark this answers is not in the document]
SukkaW: > and the matching on value instead of substr(value, length + 1, sizeof(value - 1)) part, you're extracting the prop from `key:prop;` but since we are sure that the key cannot possibly match the list of options we don't need to exclude the key from the match? That is `center|end|start|stretch` cannot possibly be in `justify-self:`.

But `start|end` can possibly be in `(flex|self)-(start|end)`. To prevent `(flex|self)-(start|end)` from being matched, it has to match the value exactly, or at least match from the start of the value. Both require strip `key` from `key:prop;`.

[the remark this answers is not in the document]
thysultan: I meant the first part namely why `match(substr(value, length + 1, sizeof(value) - 1, ...)` and not just `match(value, ...)`

thysultan:
@robinweser Kind of wish browserlist in bundlers was like eslint in so much as it would support inline comments to pieces of atomic pieces of code i.e

```
// ie11
if (a) b(c)
```
Denoting that the code only exists to make something work on ie11, then this data would make it to the bundler and strip it away from supported browser targets that are not ie11.

src/Prefixer.js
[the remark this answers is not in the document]
SukkaW: @thysultan Done in https://github.com/thysultan/stylis/pull/285/commits/999c88505871ef1e0ebc5d729976ab28e7b096db. Note checking the absence of `flex-` is still required.

SukkaW:
> The prefixer middleware, which has access to the complete state of the css token graph does not have this limitation, the `prefix` function in `Prefixer.js` is called by this middleware:

@thysultan Done in #287

src/Prefixer.js
SukkaW: I've refactored this part in https://github.com/thysultan/stylis/pull/285/commits/6adc047a7cef8db8a613b684e873131bacfd4a43

src/Prefixer.js
```diff
@@ -1,12 +1,12 @@
-import {MS, MOZ, WEBKIT} from './Enum.js'
-import {hash, charat, strlen, indexof, replace} from './Utility.js'
+import { MS, MOZ, WEBKIT } from './Enum.js'
```
thysultan, line 1: remove whitespace diffs

src/Prefixer.js
```diff
@@ -1,12 +1,12 @@
...
+import { hash, charat, strlen, indexof, replace, substr, match } from './Utility.js'
 
 /**
  * @param {string} value
  * @param {number} length
  * @return {string}
  */
-export function prefix (value, length) {
+export function prefix(value, length) {
```
thysultan, line 9: whitespace diff

src/Prefixer.js
```diff
@@ -34,10 +34,17 @@ export function prefix (value, length) {
...
+					// center, end, start, stretch
+					match(value, /(cen|end|sta|str)/) && !~indexof(value, 'flex-')
+						? MS + 'grid-row-' + replace(value, /flex-|-self/g, '')
+						: ''
+				)
+				+ value
 		// align-content
 		case 4675:
-			return WEBKIT + value + MS + 'flex-line-pack' + replace(value, /align-content|flex-|-self/, '') + value
+			return WEBKIT + value + MS + 'flex-line-pack' + replace(value, /align-content|flex-|-self/g, '') + value
```
thysultan, line 47: LGTM

src/Prefixer.js
```diff
@@ -82,12 +98,42 @@ export function prefix (value, length) {
 						return ~indexof(value, 'stretch') ? prefix(replace(value, 'stretch', 'fill-available'), length) + value : value
 				}
 			break
+		// grid-(column|row)
```
thysultan, line 101: we shouldn't use inline arrow function closure, also possibly more size optimisations, i.e hoisting the regex. I'll submit these and more to this branch this weekend.

test/Prefixer.js
```diff
@@ -186,4 +186,29 @@ describe('Prefixer', () => {
 			`animation-timing-function:cubic-bezier(0.1,0.7,1.0,0.1);`
 		].join(''))
 	})
+
+	test('grid', () => {
```
thysultan, line 190: LGTM

test/Prefixer.js
```diff
@@ -10,11 +10,11 @@ describe('Prefixer', () => {
 		expect(prefix(`flex-shrink:none;`, 11)).to.equal([`-webkit-flex-shrink:none;`, `-ms-flex-negative:none;`, `flex-shrink:none;`].join(''))
 		expect(prefix(`flex-basis:none;`, 10)).to.equal([`-webkit-flex-basis:none;`, `-ms-flex-preferred-size:none;`, `flex-basis:none;`].join(''))
 		expect(prefix(`align-self:value;`, 10)).to.equal([`-webkit-align-self:value;`, `-ms-flex-item-align:value;`, `align-self:value;`].join(''))
-		expect(prefix(`align-self:flex-start;`, 10)).to.equal([`-webkit-align-self:flex-start;`, `-ms-flex-item-align:flex-start;`, `align-self:flex-start;`].join(''))
-		expect(prefix(`align-self:flex-end;`, 10)).to.equal([`-webkit-align-self:flex-end;`, `-ms-flex-item-align:flex-end;`, `align-self:flex-end;`].join(''))
+		expect(prefix(`align-self:flex-start;`, 10)).to.equal([`-webkit-align-self:flex-start;`, `-ms-flex-item-align:start;`, `align-self:flex-start;`].join(''))
```
thysultan, line 13: still worried about theses diffs, but if PostCSS emits the same thing maybe it's not of much concern, other than that LGTM.

src/Prefixer.js
```diff
@@ -59,6 +66,15 @@ export function prefix (value, length) {
 		// justify-content
 		case 4968:
 			return replace(replace(value, /(.+:)(flex-)?(.*)/, WEBKIT + 'box-pack:$3' + MS + 'flex-pack:$3'), /s.+-b[^;]+/, 'justify') + WEBKIT + value + value
+		// justify-self
```
thysultan, line 69: TODO: size optimisations, i'll submit a patch to the branch this weekend if i get think of something to reduce.

src/Prefixer.js
```diff
@@ -34,10 +34,17 @@ export function prefix (value, length) {
 			return WEBKIT + value + replace(value, /(\w+).+(:[^]+)/, WEBKIT + 'box-$1$2' + MS + 'flex-$1$2') + value
 		// align-self
 		case 5443:
-			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/, '') + value
+			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/g, '')
```
thysultan, line 37: TODO: size optimisations. assigning to myself.

SukkaW:
@thysultan Thanks for your review! I will remove whitespace (and possibly other style-related) diff first.

test/Prefixer.js
```diff
@@ -1,4 +1,4 @@
-import {prefix} from "../index.js"
+import { prefix } from "../index.js"
```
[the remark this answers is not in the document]
SukkaW: Fixed in https://github.com/thysultan/stylis/pull/285/commits/dfae388c1aed4508b229515ff4e91fe70492f766

src/Prefixer.js
SukkaW: Fixed in https://github.com/thysultan/stylis/pull/285/commits/5f9aaf20db9f18b9110bee4850e472994a212b01

src/Prefixer.js
```diff
@@ -59,6 +59,13 @@ export function prefix (value, length) {
 		// justify-content
 		case 4968:
 			return replace(replace(value, /(.+:)(flex-)?(.*)/, WEBKIT + 'box-pack:$3' + MS + 'flex-pack:$3'), /s.+-b[^;]+/, 'justify') + WEBKIT + value + value
+		// justify-self
+		case 4200:
+			if (!match(value, /flex-|baseline/)) return MS + 'grid-column-align' + substr(value, length) + value
+			break
```
SukkaW, line 65: @thysultan 

I re-checked the implementation of `autoprefixer`, it only excludes `flex-|baseline` from being prefixed with grid:

https://github.com/postcss/autoprefixer/blob/3041c7bfced438875d668767432666b10a88ef36/lib/hacks/grid-column-align.js#L7-L9
https://github.com/postcss/autoprefixer/blob/3041c7bfced438875d668767432666b10a88ef36/lib/hacks/grid-row-align.js#L7-L9

This allows us to simplify the regexp even further, resulting in an even smaller size.

src/Prefixer.js
```diff
@@ -34,10 +34,10 @@ export function prefix (value, length) {
 			return WEBKIT + value + replace(value, /(\w+).+(:[^]+)/, WEBKIT + 'box-$1$2' + MS + 'flex-$1$2') + value
 		// align-self
 		case 5443:
-			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/, '') + value
+			return WEBKIT + value + MS + 'flex-item-' + replace(value, /flex-|-self/g, '') + (!match(value, /flex-|baseline/) ? MS + 'grid-row-' + replace(value, /flex-|-self/g, '') : '') + value
```
SukkaW, line 37: https://github.com/thysultan/stylis/pull/285/files#r824580764

thysultan: @SukkaW Any reason why `/flex-|-self/` is converted to `/flex-|-self/g` and also `/align-content|flex-|-self/g`, does a test fail because of this, can there be multiple flex- -self notations in the declaration?

src/Prefixer.js
```diff
@@ -82,12 +89,15 @@ export function prefix (value, length) {
 						return ~indexof(value, 'stretch') ? prefix(replace(value, 'stretch', 'fill-available'), length) + value : value
 				}
 			break
+		// grid-(column|row)
+		case 5152: case 5920:
+			return replace(value, /(.+?):(\d+)(\s*\/\s*(span)?\s*(\d+))?(.*)/, function (_, a, b, c, d, e, f) { return (MS + a + ':' + b + f) + (c ? (MS + a + '-span:' + (d ? e : +e - +b)) + f : '') + value })
```
thysultan, line 94: @SukkaW I updated this, can you review? does it still handle the original cases you had in mind?

SukkaW: Since the test cases passed, so I assume it does.

src/Prefixer.js
SukkaW: `align-self:flex-start;` should be prefixed as `-ms-flex-line-pack:end;`. That's why `g` is added.

thysultan:
@SukkaW Thanks for the work you've done on this.

PR #286, opened 2022-02-25
feat: add `scroll-snap` prefix support

renbaoshuo:


src/Prefixer.js
```diff
@@ -113,6 +113,12 @@ export function prefix (value, length) {
 			}
 
 			return WEBKIT + value + MS + value + value
+		// scroll-snap-type
+		case 2903:
+			return WEBKIT + value + MS + value + value
+		// scroll-margin, scroll-margin-(top|right|bottom|left)
+		case 5719: case 2647: case 2135: case 3927: case 2391:
+			return replace(value, 'scroll-margin', 'scroll-snap-margin') + value
```
SukkaW, line 121:
```suggestion
			return replace(value, 'scroll', 'scroll-snap') + value
```

src/Prefixer.js
```diff
@@ -1,12 +1,12 @@
-import {MS, MOZ, WEBKIT} from './Enum.js'
-import {hash, charat, strlen, indexof, replace} from './Utility.js'
+import { MS, MOZ, WEBKIT } from './Enum.js'
```
thysultan, line 1: whitespace diffs, here and...

src/Prefixer.js
```diff
@@ -1,12 +1,12 @@
...
+import { hash, charat, strlen, indexof, replace } from './Utility.js'
 
 /**
  * @param {string} value
  * @param {number} length
  * @return {string}
  */
-export function prefix (value, length) {
```
thysultan, line 9: whitespace diff here

PR #287, opened 2022-02-25
feat(#119): support prefix `grid-(row|column)-(start|end)`

SukkaW:
Note: Although the feature is implemented, it hasn't been optimized **AT ALL**, so suggestions, optimizations, discussions are welcome!


src/Prefixer.js
```diff
@@ -65,7 +66,24 @@ export function prefix (value, length) {
...
 		// grid-template-(columns|rows)
 		case 2592: case 3360:
-			return MS + replace(value, 'template-', '') + value
+			// TODO convert to hashes, remove arrow functions, general improvements for size
+			switch (element ? value : 0) {
+				// grid-(row|column)-start
+				case 'grid-row-start': case 'grid-column-start':
+					var end
+					// has corresponding grid-(column|row)-end
+					if (end = element && element.parent && element.parent.children && element.parent.children.find(item => item.type === DECLARATION && match(item.props, /grid-(row|column)-end/))) {
```
SukkaW, line 75: `DECLARATION` should be imported from `Enum`.

test/Middleware.js
```diff
@@ -88,5 +88,29 @@ describe('Middleware', () => {
...
+    expect(serialize(compile(`.test{grid-row-start:3;grid-row-end:5;}`), middleware([prefixer, stringify]))).to.equal([
+      `.test{-ms-grid-row:3;grid-row-start:3;-ms-grid-row-span:2;grid-row-end:5;}`
+    ].join(''))
+    // should not prefix a cell with non-numerical position values
+    expect(serialize(compile(`.test{grid-row-start:3;grid-row-end:span 5;}`), middleware([prefixer, stringify]))).to.equal([
+      `.test{grid-row-start:3;grid-row-end:span 5;}`
+    ].join(''))
+    expect(serialize(compile(`.test{grid-row-start:span 3;grid-row-end:5;}`), middleware([prefixer, stringify]))).to.equal([
+      `.test{grid-row-start:span 3;grid-row-end:5;}`
+    ].join(''))
```
SukkaW, line 114: We should move this part to the `test/Prefixer`, under `grid`.

thysultan: This should actually still be in middleware.

SukkaW: If the feature is now being implemented in the `Prefixer`, the test case should be moved, too.

thysultan: Ok, it should however still be the middleware variant it was before, syntax wise.

src/Prefixer.js
```diff
@@ -66,6 +67,16 @@ export function prefix (value, length) {
 		// grid-template-(columns|rows)
 		case 2592: case 3360:
 			return MS + replace(value, 'template-', '') + value
+		// grid-(row|column)-start
+		case 4384: case 3616:
+			if (children = children && children.find(function (element) { return match(element.props, /grid-\w+-end/) })) {
```
thysultan, line 72: [IE11 doesn't support Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find). Would defeat the purpose to support IE11 grids with unsupported JS api's. Let's use `.some` and set the length variable when a match is found we use that to index into children i.e

```js
if (children.some(function (element, index) { return length = index, match(element.props, /grid-\w+-end/) })) {
```

Then below just:

```js
return ~indexof(value + (children = children[length].value), 'span')
```

thysultan:
LGTM, @SukkaW Thanks for both of these, hopefully will release later today if nothing else come up.

PR #289, opened 2022-04-21
fix: support `value !important;`

SukkaW:
Fix an edge case that `position:sticky!important` can be prefixed while `position:sticky !important` (a white space before `!important`) can not.

### DIFF
--- a/src/Prefixer.js
+++ b/src/Prefixer.js
@@ -105,20 +105,18 @@ export function prefix (value, length, children) {
 			return replace(value, /(.+?):(\d+)(\s*\/\s*(span)?\s*(\d+))?(.*)/, function (_, a, b, c, d, e, f) { return (MS + a + ':' + b + f) + (c ? (MS + a + '-span:' + (d ? e : +e - +b)) + f : '') + value })
 		// position: sticky
 		case 4949:
-			// (s)ticky?
-			if (charat(value, length + 1) !== 115)
-				break
+			// stick(y)?
+			if (charat(value, length + 6) === 121)
+				return replace(value, ':', ':' + WEBKIT) + value
+			break
 		// display: (flex|inline-flex|grid|inline-grid)
 		case 6444:
-			switch (charat(value, strlen(value) - 3 - (~indexof(value, ' !important') ? 11 : ~indexof(value, '!important') && 10))) {
-				// stic(k)y
-				case 107:
-					return replace(value, ':', ':' + WEBKIT) + value
-				// (inline-)?fl(e)x
-				case 101:
-					return replace(value, /(.+:)([^;\s!]+)(;|\s?!.+)?/, '$1' + WEBKIT + (charat(value, 14) === 45 ? 'inline-' : '') + 'box$3' + '$1' + WEBKIT + '$2$3' + '$1' + MS + '$2box$3') + value
-				// (inline-)?grid
-				case 105:
+			switch (charat(value, 14) === 45 ? charat(value, 18) : charat(value, 11)) {
+				// (inline-)?fle(x)
+				case 120:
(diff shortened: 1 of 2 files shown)


src/Prefixer.js
```diff
@@ -110,13 +110,13 @@ export function prefix (value, length, children) {
 				break
 		// display: (flex|inline-flex|grid|inline-grid)
 		case 6444:
-			switch (charat(value, strlen(value) - 3 - (~indexof(value, '!important') && 10))) {
+			switch (charat(value, strlen(value) - 3 - (~indexof(value, ' !important') ? 11 : ~indexof(value, '!important') && 10))) {
```
Andarist, line 113: this probably won't work for `value          !important`

SukkaW: Fixed. And related test cases are added.

src/Prefixer.js
```diff
@@ -105,20 +105,18 @@ export function prefix (value, length, children) {
...
+			// stick(y)?
+			if (charat(value, length + 6) === 121)
+				return replace(value, ':', ':' + WEBKIT) + value
+			break
 		// display: (flex|inline-flex|grid|inline-grid)
 		case 6444:
-			switch (charat(value, strlen(value) - 3 - (~indexof(value, '!important') && 10))) {
-				// stic(k)y
-				case 107:
-					return replace(value, ':', ':' + WEBKIT) + value
```
SukkaW, line 116: Instead of letting `position: sticky` fallthrough to `display` and handle it there, the change now handles it in `position` directly, and either break or return prefixed to stop it from being fallthrough.

src/Prefixer.js
```diff
@@ -105,20 +105,18 @@ export function prefix (value, length, children) {
...
-				// (inline-)?fl(e)x
-				case 101:
-					return replace(value, /(.+:)([^;!]+)(;|!.+)?/, '$1' + WEBKIT + (charat(value, 14) === 45 ? 'inline-' : '') + 'box$3' + '$1' + WEBKIT + '$2$3' + '$1' + MS + '$2box$3') + value
-				// (inline-)?grid
-				case 105:
+			switch (charat(value, 14) === 45 ? charat(value, 18) : charat(value, 11)) {
+				// (inline-)?fle(x)
+				case 120:
+					return replace(value, /(.+:)([^;\s!]+)(;|(\s+)?!.+)?/, '$1' + WEBKIT + (charat(value, 14) === 45 ? 'inline-' : '') + 'box$3' + '$1' + WEBKIT + '$2$3' + '$1' + MS + '$2box$3') + value
```
SukkaW, line 117:
```
display:(f)lex
display:(f)low

display:f(l)ex
display:f(l)ow
display:b(l)ock

display:gr(i)d
display:in(i)tial
```

So the fourth letter is being used.

test/Prefixer.js
```diff
@@ -1,9 +1,14 @@
 import {compile, serialize, stringify, middleware, prefixer, prefix} from "../index.js"
 
+const globalCssValues = ['inherit', 'initial', 'unset', 'revert', 'revert-layer']
```
SukkaW, line 3: https://github.com/thysultan/stylis/pull/289/files#r854961406

thysultan: Can inline it in the single test case.

SukkaW: @thysultan  `globalCssValues` is being used twice (That's why I extracted it in the first place). One for `display` and one for `position`. And in the future PR I will add more test cases related with it.

src/Prefixer.js
SukkaW: Related test cases have been added.

thysultan:
As always, thanks: published 4.1.1 with both of [these](https://github.com/thysultan/stylis/pull/288).

PR #298, opened 2022-09-02
Update Parser.js

sam007mac:
Functionality is the same, but previous version would break when minified with the Google Closure Compiler.

(it ended up doing the equivalent of `var y = z = substr(...)`, therefore the if statement would always return false)


Andarist:
> Functionality is the same, but previous version would break when minified with the Google Closure Compiler.

It's probably best to report this issue to that project.

PR #302, opened 2022-09-09
Avoid prefixing css vars

Andarist:
fixes https://github.com/thysultan/stylis/issues/301

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -1,5 +1,5 @@
 import {COMMENT, RULESET, DECLARATION} from './Enum.js'
-import {abs, trim, from, sizeof, strlen, substr, append, replace, indexof} from './Utility.js'
+import {abs, charat, trim, from, sizeof, strlen, substr, append, replace, indexof} from './Utility.js'
 import {node, char, prev, next, peek, caret, alloc, dealloc, delimit, whitespace, escaping, identifier, commenter} from './Tokenizer.js'
 
 /**
@@ -43,7 +43,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 		switch (previous = character, character = next()) {
 			// (
 			case 40:
-				if (previous != 108 && characters.charCodeAt(length - 1) == 58) {
+				if (previous != 108 && charat(characters, length - 1) == 58) {
 					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f') != -1)
 						ampersand = -1
 					break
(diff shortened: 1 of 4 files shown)


src/Parser.js
```diff
@@ -43,7 +43,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 		switch (previous = character, character = next()) {
 			// (
 			case 40:
-				if (previous != 108 && characters.charCodeAt(length - 1) == 58) {
+				if (previous != 108 && charat(characters, length - 1) == 58) {
```
Andarist, line 46: This is just a bonus change, not rly relevant to this PR

src/Prefixer.js
```diff
@@ -8,7 +8,7 @@ import {hash, charat, strlen, indexof, replace, substr, match} from './Utility.j
  * @return {string}
  */
 export function prefix (value, length, children) {
-	switch (hash(value, length)) {
+	switch (charat(value, 0) != 45 && charat(value, 1) != 45 && hash(value, length)) {
```
Andarist, line 11: the expression's type now is `false | number` - perhaps we should convert `false` to a `0`? would that have any effect on the engine? According to some TS PRs that I've seen lately (like [this one](https://github.com/microsoft/TypeScript/pull/50225)) JS engines don't do jump-table optimization for switch statements anyway so perhaps this doesn't matter at all? I'm not sure 🤷‍♂️

test/Prefixer.js
```diff
@@ -271,4 +271,8 @@ describe('Prefixer', () => {
 	test('tab-size', () => {
 		expect(prefix(`tab-size:1;`, 8)).to.equal([`-moz-tab-size:1;`, `tab-size:1;`].join(''))
 	})
+
+	test('css variables', () => {
+		expect(prefix(`--CircularProgress-animation:0.5s linear;`, 28)).to.equal('--CircularProgress-animation:0.5s linear;')
```
Andarist, line 276: Without this change we were accidentally hitting this branch of the code for this input:
https://github.com/thysultan/stylis/blob/08b179da63fe6f06ffc2ca9a166bbc5003366c33/src/Prefixer.js#L21-L23

src/Prefixer.js
thysultan: The second `charat(value, 0) != 45` is probably not needed as we also prefer to not touch `-webkit-` already prefixed props. 

It'd put `charat(value, 0) ^ 45 ? 0 : ...` inside the `hash` function.

Andarist: Done

PR #304, opened 2022-10-25
Add support for container queries

### LINKED ISSUE #303: Support for @container
Chrome 105 shipped with support for container query, would be great if this would be supported.
Example:
```css
@container (max-width: 300px) {
  .element {
    display: none;
  }
}
```

lunaris:
Fixes #303. Closes #300.

This PR adds support for container queries, parsing and hoisting them in the same manner that media queries are. I'm not sure of the best way to do this in a manner that yields acceptable performance/conforms to the existing patterns in stylis, but I've taken a stab. The challenge is that currently, we only need to look ahead one character after an `@` to determine what to do, but with `@container` this is not sufficient as there is also `@counter-style`. I've thus taken what feels like a hacky approach to look ahead until the `n`, which I think is safe and seems to satisfy the test suite. All feedback gratefully received though!

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -93,7 +93,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 							if (offset === 0)
 								parse(characters, root, reference, reference, props, rulesets, length, points, children)
 							else
-								switch (atrule) {
+								switch (atrule === 99 && charat(characters, 3) === 110 ? 100 : atrule) {
 									// d m s
 									case 100: case 109: case 115:
 										parse(value, reference, reference, rule && append(ruleset(value, reference, reference, 0, 0, rules, points, type, rules, props = [], length), children), rules, children, length, points, rule ? props : children)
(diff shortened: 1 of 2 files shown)


src/Parser.js
```diff
@@ -98,6 +98,13 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 									case 100: case 109: case 115:
 										parse(value, reference, reference, rule && append(ruleset(value, reference, reference, 0, 0, rules, points, type, rules, props = [], length), children), rules, children, length, points, rule ? props : children)
 										break
+									// c
```
thysultan, line 101:
```diff
+ switch (atrule === 99 && charat(characters, 3) === 110 ? 100 : atrule) {
- switch (atrule) {
```

lunaris: Is this a change you'd like me to make? Not sure if a commit or other suggestion has gone missing here. Also I'm not sure if there's a benchmark or the like you'd like me to run beforehand? Thanks for reviewing!

thysultan: Yes, this, instead of the current implementation, this one uses the least bytes and branches.

PR #307, opened 2022-11-17
Add a failing test case for `&` within a nested function in a declaration value

Andarist:
Related to https://github.com/emotion-js/emotion/issues/2943

### DIFF
--- a/test/Parser.js
+++ b/test/Parser.js
@@ -122,6 +122,16 @@ describe('Parser', () => {
     ).to.equal(`.user{color:red;}`)
   })
 
+  test('& in value in a nested function', () => {
+    const urlOneX = "https://images.ctfassets.net/test.jpg?fm=webp&q=70&w=1000";
+    const urlTwoX = "https://images.ctfassets.net/test.jpg?fm=webp&q=70&w=2000";
+    expect(
+      stylis(`
+        background-image: image-set(url(${urlOneX}) 1x, url(${urlTwoX}) 2x);
+     `)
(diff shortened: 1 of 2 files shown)


test/Parser.js
```diff
@@ -122,6 +122,16 @@ describe('Parser', () => {
...
   })
 
+  test('& in value in a nested function', () => {
+    const urlOneX = "https://images.ctfassets.net/test.jpg?fm=webp&q=70&w=1000";
+    const urlTwoX = "https://images.ctfassets.net/test.jpg?fm=webp&q=70&w=2000";
+    expect(
+      stylis(`
+        background-image: image-set(url(${urlOneX}) 1x, url(${urlTwoX}) 2x);
+     `)
+    ).to.equal(`.user{background-image:image-set(url(${urlOneX}) 1x, url(${urlTwoX}) 2x);}`)
```
Andarist, line 132: `\f` gets incorrectly appended to `&` here:
https://github.com/thysultan/stylis/blob/bc0b1b03f5562eb8b38601e4e05461631a91c400/src/Parser.js#L47-L49

I'd like to conditionally use this `replace` here - we need to skip it when we are in the context of the property. I've tried to use `property` for this and while I managed to fix this case I broke some other cases like this one:
https://github.com/thysultan/stylis/blob/bc0b1b03f5562eb8b38601e4e05461631a91c400/test/Parser.js#L593-L601

Andarist: @thysultan friendly 🏓

thysultan:
@Andarist didn't notice this when you opened it, (only the friendly ping): fixed in the head commit, you can merge/release this one.

thysultan (approved):

PR #309, opened 2022-12-30
fix: webkit mask-composite values

Olivia-SY129:
Fix for mask-composite prop values for webkit

Here are the changes:
```
'add' → 'source-over'
'substract' → 'source-out'
'intersect' → 'source-in'
'exclude' → 'xor'
```

Example for ```element.return```:
```
before fix (not working in webkit)
'-webkit-mask-composite: add,intersect;mask-composite: add,intersect;'  

after fix
'-webkit-mask-composite: source-over,source-in;mask-composite: add,intersect;'
```


thysultan:
@Olivia-SY129 can you add a test case, i'm assuming we can replace `replaceAll` with just `replace` safe assumption?

thysultan (approved):

PR #313, opened 2023-05-04
feat: support @layer queries

probablyup:
I'd be lying if I told you I knew what this code was doing, but it appears to work.

closes https://github.com/thysultan/stylis/issues/312


src/Parser.js
```diff
@@ -94,8 +94,8 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 								parse(characters, root, reference, reference, props, rulesets, length, points, children)
 							else
 								switch (atrule === 99 && charat(characters, 3) === 110 ? 100 : atrule) {
-									// d m s
-									case 100: case 109: case 115:
+									// d m s l
+									case 100: case 109: case 115: case 108:
```
Andarist, line 98:
```suggestion
									// d l m s
									case 100: case 108: case 109: case 115:
```

src/Parser.js
```diff
@@ -94,8 +94,8 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 								parse(characters, root, reference, reference, props, rulesets, length, points, children)
 							else
 								switch (atrule === 99 && charat(characters, 3) === 110 ? 100 : atrule) {
-									// d m s
-									case 100: case 109: case 115:
+									// d l m s
```
Andarist, line 97: for completeness you should also add `export var LAYER = '@layer'` here:
https://github.com/thysultan/stylis/blob/ac5e81d7253c3d97e7121cba5bd36530c286ca55/src/Enum.js#L9-L20

probablyup:
@Andarist added most of what was discussed in the PR

test/Parser.js
```diff
@@ -1111,4 +1111,14 @@ describe('Parser', () => {
       `.user.foo.bar.barbar{color:orange;}`
     ].join(''))
   })
+
+  test('@layer query (#312)', () => {
```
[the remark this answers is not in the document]
probablyup: > Can @layer be nested in @media? Or is it only allowed to nest @media in @layer?

Based on [this article](https://www.bram.us/2021/09/15/the-future-of-css-cascade-layers-css-at-layer/#more-notes--conditionals), it appears `@layer` and `@media` can be nested interchangeably but there is significance in the order they're nested in terms of browser evaluation.

Andarist (approved):

test/Parser.js
```diff
@@ -1111,4 +1111,105 @@ describe('Parser', () => {
...
```
Andarist, line 1115:
```suggestion
  test('cascade @layer (#312)', () => {
```

PR #315, opened 2023-06-04
chore: add failing test for prefixer + rulesheet re: placeholder pseudo

probablyup:
Based on my understanding of `rulesheet`, I think the `@media {::placeholder{}}` combo should be split into separate rules.

### DIFF
--- a/test/Middleware.js
+++ b/test/Middleware.js
@@ -4,7 +4,7 @@ const stack = []
 
 describe('Middleware', () => {
 	test('rulesheet', () => {
-  	serialize(compile(`@import url('something.com/file.css');.user{ h1 {width:0;} @media{width:1;}@keyframes name{from{width:0;}to{width:1;}}}@keyframes empty{}`), middleware([prefixer, stringify, rulesheet(value => stack.push(value))]))
+  	serialize(compile(`@import url('something.com/file.css');.user{ h1 {width:0;} @media{width:1;}@keyframes name{from{width:0;}to{width:1;}}}@keyframes empty{}@media (min-width: 500px){.a::placeholder{color: red;}}`), middleware([prefixer, stringify, rulesheet(value => stack.push(value))]))
   	expect(stack).to.deep.equal([
       `@import url('something.com/file.css');`,
       `.user h1{width:0;}`,
@@ -13,6 +13,10 @@ describe('Middleware', () => {
       `@keyframes name{from{width:0;}to{width:1;}}`,
       '@-webkit-keyframes empty{}',
       '@keyframes empty{}',
+      '@media (min-width: 500px){.a::-webkit-input-placeholder{color:red;}}',
+      '@media (min-width: 500px){.a::-moz-placeholder{color:red;}}',
+      '@media (min-width: 500px){.a:-ms-input-placeholder{color:red;}}',
+      '@media (min-width: 500px){.a::placeholder{color:red;}}',
     ])
   })
 


probablyup:
@thysultan if you have any pointers for solving this, I'd be happy to try. Just haven't had much luck so far.

Andarist:
I **think** that the fix would have to be applied [here](https://github.com/thysultan/stylis/blob/1bf92ea2338610a49e3e0ae06d9e6ed8f6f986a9/src/Middleware.js#L51-L67 and *somehow* you'd have to recursively copy the parent rules through their `.root` (if there are any parent rules) and put that single prefixed rule as the "leaf" `.children`

Andarist:
Almost managed to fix it - there is just one failing test case:
https://github.com/thysultan/stylis/blob/5687d854384bb0285130581e951a1d620128f63f/test/Middleware.js#L84-L87

This test tests the **output** and not what is passed to `rulesheet` (`rulesheet` is just a plugin), values passed to `rulesheet` (I tested this manually) are OK.

I'm not sure how to resolve this conflict because prefixing of those pseudo rules happens transitively here:
https://github.com/thysultan/stylis/blob/5687d854384bb0285130581e951a1d620128f63f/src/Serializer.js#L35

As we can see - this serializes `children` and uses that as part of the value. This happens when stringifying `@media` rule and the serialized `children` land within `{ ... }`. The problem is that if we return the prefixed stuff here:
https://github.com/thysultan/stylis/blob/5687d854384bb0285130581e951a1d620128f63f/src/Middleware.js#L57
then we end up with outputs like:
```css
@media (min-width: 500px){a:-moz-read-only{color:red;}}
@media (min-width: 500px){@media (min-width: 500px){a:-moz-read-only{color:red;}}a:read-only{color:red;}}
```

We can see here that the prefixed and stringified top-level rule becomes part of the children of that top-level rule which is not what we want to achieve here.

This problem originates in the assumption that the prefixed things always become siblings of what they "cloned" but that's no longer true. 

cc @thysultan

probablyup:
Ping @thysultan, could you spare some time to look at this one? Thank you!

thysultan:
Yes, i'll take a look at this, this weekend.

thysultan:
@Andarist So the latest commit fixes the issue, but the order of rules are: changed to `unprefixed` then `prefix` for `::placeholder`.

thysultan:
Ok, i think i managed to preserve the `prefix-first` order now.

probablyup:
Thank you @thysultan, your code wizardry is appreciated as always

joacub:
this pullrequest introduce an error that makes push not been an array anymore.

joacub:
this changes cause this error:

image

Andarist:
Please share a repro case

HelloWorld017:
I think that if we use the `prefixer` from `stylis@4.3.0` with `@emotion/react@11.10.0`, it would break, because nodes parsed with older version of the stylis will not have the siblings property.

For those who have the same issue:  I have mitigated this by downgrading to `stylis@4.2.0`.


<details>
<summary>Keywords & Stacktraces for issue searchers:</summary>

Keywords:  
`::placeholder` `stylis` `emotion` `push`  `undefined`

Stacktraces:
```
TypeError: Cannot read properties of undefined (reading 'push')
    at append (file:///node_modules/stylis/src/Utility.js:105:15)
    at lift (file:///node_modules/stylis/src/Tokenizer.js:40:2)
    at file:///node_modules/stylis/src/Middleware.js:62:10
    at Array.map (<anonymous>)
    at combine (file:///node_modules/stylis/src/Utility.js:114:15)
    at Array.i (file:///node_modules/stylis/src/Middleware.js:52:14
```

</details>

PR #316, opened 2023-07-16
Update Tokenizer.js

mohsenomidi:
Fix Can not read properties undefined (reading 'push')

### DIFF
--- a/src/Tokenizer.js
+++ b/src/Tokenizer.js
@@ -37,7 +37,7 @@ export function lift (root) {
 	while (root.root)
 		root = copy(root.root, {children: [root]})
 
-	append(root, root.siblings)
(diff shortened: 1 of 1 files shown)


mohsenomidi:
Fix [mui - issue 37900](https://github.com/mui/material-ui/issues/37900)

Andarist:
This feels like a bandaid for some deeper problem. We need to diagnose the deeper problem and we need a test case to be added to the fixing PR, one that would fail without the proposed patch.

mohsenomidi:
Thanks for the review, then we are waiting to find out the root cause and fix.

CodingItWrong:
I experienced this error as well on 4.3.0. As a note to others, downgrading stylis to 4.2.0 fixed it for me

PR #322, opened 2023-12-22
Fixed `&` search after consuming parenthesis

Andarist:
fixes https://github.com/thysultan/stylis/issues/321

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -44,7 +44,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 			// (
 			case 40:
 				if (charat(characters, length - 1) == 58) {
-					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f', abs(points[index - 1])) != -1)
+					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f', abs(index ? points[index - 1] : 0)) != -1)
 						ampersand = -1
 					break
 				}
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -43,8 +43,8 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 		switch (previous = character, character = next()) {
 			// (
 			case 40:
-				if (previous != 108 && charat(characters, length - 1) == 58) {
-					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f') != -1)
+				if (charat(characters, length - 1) == 58) {
+					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f', abs(points[index - 1])) != -1)
 						ampersand = -1
 					break
 				}
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -43,7 +43,7 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 		switch (previous = character, character = next()) {
 			// (
 			case 40:
-				if (charat(characters, length - 1) == 58) {
+				if (previous != 108 && charat(characters, length - 1) == 58) {
 					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f', abs(index ? points[index - 1] : 0)) != -1)
 						ampersand = -1
 					break
--- a/src/Utility.js
+++ b/src/Utility.js
@@ -55,10 +55,11 @@ export function replace (value, pattern, replacement) {
 /**
  * @param {string} value
  * @param {string} search
+ * @param {number} position
  * @return {number}
  */
-export function indexof (value, search) {
-	return value.indexOf(search)
+export function indexof (value, search, position) {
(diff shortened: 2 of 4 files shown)


src/Parser.js
```diff
@@ -43,8 +43,8 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 		switch (previous = character, character = next()) {
 			// (
 			case 40:
-				if (previous != 108 && charat(characters, length - 1) == 58) {
```
Andarist, line 46: I have no idea what this `previous` check was doing here. `108` is `l` and I don't see any specific cases related to that. All tests pass too.

src/Parser.js
```diff
@@ -43,8 +43,8 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
...
-					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f') != -1)
+				if (charat(characters, length - 1) == 58) {
+					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f', abs(points[index - 1])) != -1)
```
Andarist, line 47: iiuc this should limit the search to the portion of the text that contains the current selector. One potential problem I see here is that it might result in out-of-bounds access on `points`. Is this something to worry about at all from the engines' PoV?

src/Utility.js
```diff
@@ -55,10 +55,11 @@ export function replace (value, pattern, replacement) {
 /**
  * @param {string} value
  * @param {string} search
+ * @param {number} position
  * @return {number}
  */
-export function indexof (value, search) {
-	return value.indexOf(search)
+export function indexof (value, search, position) {
```
Andarist, line 61: perhaps now this function stops being homomorphic, should I pass `0` as an extra argument at all other call sites? 🤔

src/Parser.js
thysultan: it's trying to avoid both `:global(` and `url(`

Andarist: would you mind adding a test case for this?

src/Utility.js
thysultan: Yes

src/Parser.js
thysultan: yes, out of bound access would be slower in the js engine

```js
// (
			case 40:
				if (previous != 108 && charat(characters, length - 1) == 58) {
					if (indexof(characters += replace(delimit(character), '&', '&\f'), '&\f', abs(index ? points[index - 1] : 0)) != -1)
						ampersand = -1
					break
				}
```

src/Parser.js
thysultan:
```js
test.only('context character VIIII', () => {
    expect(
      stylis(`background: url(i&m&g.png);.a {background: url(i&m&g.png);}`)
    ).to.equal([
      `.user{background:url(i&m&g.png);}`,
      `.user .a{background:url(i&m&g.png);}`
    ].join(''))
  })
```

but we already have a stop gap in 

```js
case 59 + offset: if (ampersand == -1) characters = replace(characters, /\f/g, '')
```

that means that test will pass either way, but if you remove the 

```js
if (ampersand == -1) characters = replace(characters, /\f/g, '')
``` 

you can see it in action, so the check for `l` is an early exit before that.

Andarist: so is it merely an optimization here?

src/Parser.js
Andarist: > also garbage data in points array if it's not in a comma separated selector so we should check if it is before.

I don't quite understand this. My understanding of this parser is quite superficial ;p especially with such long breaks between my small tweaks here and there. I'm not sure in what situation some garbage could end up in `points` here.

Andarist:
@thysultan would you mind taking another look at this?

src/Parser.js
thysultan: Yes

src/Parser.js
thysultan: `points` is never cleared/zero'ed out between function calls, it act's like a reusable stack(we avoid growing it if need be), as such reading any random index(in this case zero) into it could give you some random value from a previous stack call if you aren't sure that `points` is being used in the current stack call, the `points[index++] = ...`

src/Parser.js
Andarist: ok, i brought this back

src/Parser.js
Andarist: ok, it makes more sense now

Andarist:
I think this is ready to be merged in now - but a final confirmation would be highly appreciated :)

quantizor:
@Andarist could you cut a test build so I can cross-check?

Andarist:
@quantizor I don't quite want to go through this hassle right now. Would you be willing to try this out through `npm link` or similar?

thysultan (approved):

quantizor:
Thank you both!

PR #329, opened 2024-04-23
Add @scope enum

typicode:
Adds @scope to the list of enums
https://developer.mozilla.org/en-US/docs/Web/CSS/@scope


typicode:
Thanks for the quick merge and the project!

Andarist:
Thanks for husky 😉

typicode:
😁

PR #339, opened 2024-12-03
test: add more unit test

Wxh16144:
I'd like to report an issue (though I'm not entirely sure whether it's a bug or just user error).

I found an issue report on https://github.com/ant-design/ant-design/issues/51867.

After investigating, I traced it down to `stylis` having problems with parsing the `@layer` query.

Here's a minimal example:

```jsx
import { compile, serialize, stringify } from "stylis"

serialize(compile(`@layer base{}`), stringify) // should return `@layer base{}`
```

The expected output is `@layer base{}`, but it actually returns `@layer base`, missing the `{}` part.

I also tested `@emotion/styled`, and it works as expected.

image

It seems like an edge case, and I believe it's a bug.

Thanks for your time. I hope this gets resolved soon.

### DIFF
--- a/test/Parser.js
+++ b/test/Parser.js
@@ -1259,5 +1259,11 @@ describe('Parser', () => {
         }
       `)
     ).to.equal(`@layer framework{@keyframes slide-left{from{margin-left:0;}to{margin-left:-100%;}}}`);
+
+  });
+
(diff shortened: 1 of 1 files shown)


Wxh16144:
Are there any updates? If this is as expected, please close it in a timely manner.

PR #342, opened 2025-01-12
Fixed `@left-*` parsing

Andarist:
fixes https://github.com/thysultan/stylis/issues/341

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -97,16 +97,14 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
 								switch (atrule) {
 									// c(ontainer)
 									case 99:
-										if (charat(characters, 3) !== 110) offset = 0
-										break
+										if (charat(characters, 3) === 110) break
 									// l(ayer)
 									case 108:
-										if (charat(characters, 2) !== 97) offset = 0
-										break
+										if (charat(characters, 2) === 97) break
+									default:
+										offset = 0
 									// d(ocument) m(edia) s(upports)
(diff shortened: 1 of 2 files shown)


src/Parser.js
```diff
@@ -93,15 +93,22 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
...
-										break
+							else {
+								switch (atrule) {
+									// c(ontainer)
+									case 99:
+										if (charat(characters, 3) === 110) break
+									// l(ayer)
+									case 108:
+										if (charat(characters, 2) === 97) break
 									default:
```
Andarist, line 104: this is a default case that's not the last one 😅 I tried to code-golf this as much as possible and this is my best version so far

I was testing the size increase on `dist/umd/stylis.js` (since it's already minified) using `wc -c` and `gzip -c | wc -c`. Those are results:
| ref     | wc -c | gzip -c \| wc -c |
|---------|-------|------------------|
| master  | 10571 | 4396             |
| 9c57efb | 10607 | 4413             |
| 2a519d3 | 10619 | 4409             |
| ebd509a | 10606 | 4406             |

Andarist: The idea behind the fix here is that for non-layer and non-container (this case was already covered) we "reset" `offset`. We do the same in the default case and that's the code falls through from those when it doesn't `break` into that non-last default

Then based on the `offset` (if it survived or not) the correct parsing invocation is selected

thysultan (approved):

PR #353, opened 2026-03-19
Fix `&` substitutions in functional pseudo classes

### LINKED ISSUE #350: Multiple `&` references inside `:is()`, `:where()`, `:has()` not fully resolved
## Description

When multiple `&` (parent selector) references appear inside `:is()`, `:where()`, or `:has()` pseudo-class arguments, stylis only resolves the first `&` but leaves subsequent ones as literal `&` characters.

## Reproduction

```js
import { compile, serialize, stringify } from 'stylis';

const result = compile('.parent .child { :is(&:hover, .other:hover &) .target { color: red; } }');
console.log(serialize(result, stringify));
```

### Expected output
```css
:is(.parent .child:hover, .other:hover .parent .child) .target{color:red;}
```

### Actual output
```css
:is(.parent .child:hover, .other:hover &) .target{color:red;}
```

The second `&` in the `:is()` arguments is not replaced with the parent selector.

## Environment

stylis 4.3.6

Andarist:
fixes https://github.com/thysultan/stylis/issues/350

### DIFF
--- a/src/Parser.js
+++ b/src/Parser.js
@@ -1,5 +1,5 @@
 import {COMMENT, RULESET, DECLARATION} from './Enum.js'
-import {abs, charat, trim, from, sizeof, strlen, substr, append, replace, indexof} from './Utility.js'
+import {abs, charat, trim, from, sizeof, strlen, substr, append, replace} from './Utility.js'
 import {node, char, prev, next, peek, token, caret, alloc, dealloc, delimit, whitespace, escaping, identifier, commenter} from './Tokenizer.js'
 
 /**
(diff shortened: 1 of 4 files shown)


src/Parser.js
```diff
@@ -43,17 +44,23 @@ export function parse (value, root, parent, rule, rules, rulesets, pseudo, point
...
 			// " ' [
 			case 34: case 39: case 91:
 				characters += delimit(character)
 				break
 			// \t \n \r \s
 			case 9: case 10: case 13: case 32:
+				if (parens > 0) {
+					characters += from(character)
+					break
+				}
```
Andarist, line 63: this one is potentially redundant but I erred on the side of keeping the current parsing output in regards to the whitespace within the parenthesis

I couldn't quite find a reason to justify whitespace preservation in this context but I vaguely recall that was made on purpose (I can't find an issue reference or anything like that though)

thysultan (approved):

